Issue #29, opened 2019-01-20
GetPage fails with error : 'Cannot convert array to rectangle'


jdenieul:
Hello,

When i try to open a PDF file and read it, i have an error : 

UglyToad.PdfPig.Exceptions.PdfDocumentFormatException : 'Cannot convert array to rectangle, expected 4 values instead got: [ 0, 0 ].'

UglyToad.PdfPig.Exceptions.PdfDocumentFormatException
  HResult=0x80131500
  Message=Cannot convert array to rectangle, expected 4 values instead got: [ 0, 0 ].
  Source=UglyToad.PdfPig
  Arborescence des appels de procédure :
   à UglyToad.PdfPig.Util.ArrayTokenExtensions.ToIntRectangle(ArrayToken array)
   à UglyToad.PdfPig.Parser.PageFactory.GetMediaBox(Int32 number, DictionaryToken dictionary, PageTreeMembers pageTreeMembers, Boolean isLenientParsing)
   à UglyToad.PdfPig.Parser.PageFactory.Create(Int32 number, DictionaryToken dictionary, PageTreeMembers pageTreeMembers, Boolean isLenientParsing)
   à UglyToad.PdfPig.Content.Pages.GetPage(Int32 pageNumber)
   à UglyToad.PdfPig.PdfDocument.GetPage(Int32 pageNumber)
   à WindowsFormsApp2.Form1.button1_Click(Object sender, EventArgs e) dans C:\Users\source\repos\WindowsFormsApp2\WindowsFormsApp2\Form1.cs :ligne 40
   à System.Windows.Forms.Control.OnClick(EventArgs e)
   à System.Windows.Forms.Button.OnClick(EventArgs e)
   à System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   à System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   à System.Windows.Forms.Control.WndProc(Message& m)
   à System.Windows.Forms.ButtonBase.WndProc(Message& m)
   à System.Windows.Forms.Button.WndProc(Message& m)
   à System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message& m)
   à System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   à System.Windows.Forms.NativeWindow.DebuggableCallback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
[...]

PR #30, opened 2019-04-19
Fix issue with Width, Height and Area

BobLd:
Fix issue with Width, Height and Area when the text orientation is not horizontal.

### DIFF
--- a/src/UglyToad.PdfPig/Content/Letter.cs
+++ b/src/UglyToad.PdfPig/Content/Letter.cs
@@ -72,7 +72,7 @@
             Width = width;
             StartBaseLine = startBaseLine;
             EndBaseLine = endBaseLine;
-            TextDirection = getTextDirection();
+            TextDirection = GetTextDirection();
         }
 
         /// <summary>
@@ -83,7 +83,7 @@
             return $"{Value} {Location} {FontName} {PointSize}";
         }
 
-        private TextDirection getTextDirection()
+        private TextDirection GetTextDirection()
         {
             if (System.Math.Abs(StartBaseLine.Y - EndBaseLine.Y) < 10e-5m)
             {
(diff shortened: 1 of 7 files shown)


EliotJones:
Thanks so much for this, I'm probably not going to get time to take a proper look this weekend but I'll try to get it reviewed next week.

Thanks again.

BobLd:
Hey Eliot, no worries. I intend to implement a IWordExtractor that leverages document layout analysis.
Having accurate bounding boxes will help a lot ;)

src/UglyToad.PdfPig/Content/Letter.cs
```diff
@@ -65,5 +82,26 @@ public override string ToString()
         {
             return $"{Value} {Location} {FontName} {PointSize}";
         }
+
+        private TextDirection getTextDirection()
```
EliotJones: Just a minor style thing, this should be `GetTextDirection` even for private methods.

src/UglyToad.PdfPig/Geometry/PdfRectangle.cs
```diff
@@ -33,17 +33,17 @@ public struct PdfRectangle
...
         /// Width of the rectangle.
         /// </summary>
-        public decimal Width { get; }
+        public decimal Width => Right - Left;
 
         /// <summary>
         /// Height of the rectangle.
         /// </summary>
-        public decimal Height { get; }
+        public decimal Height => Top - Bottom;
```
EliotJones: While this is an improvement because it prevents repetition of this logic in each constructor it also means these calculations are repeated for every property access. I can't think of a scenario yet where that could cause issues but it's worth bearing in mind though I don't think it needs to be changed back.

BobLd:
Thanks for the feedback!

EliotJones:
Thanks for making that change, I've merged this in now.

Do you need the functionality in a Nugettable release in the near future? I think the code for 0.0.6 is still potentially broken with the TypeInitialiser error mentioned here in SystemFontFinder: https://github.com/UglyToad/PdfPig/issues/29#issuecomment-457562934 but I've not been able to reproduce that issue (it's OS dependent) so I might do a soft pre-release release of the 0.0.6 functionality and see if people start reporting the same problem.

PR #31, opened 2019-05-12
Adding 'Line' and 'PdfLine'

BobLd:
In Content:
  - Adding a 'Line' of text object
  - Adding a 'TextDirection' property in the 'Word' object

In Geometry:
  - Adding a 'PdfLine' object
  - Making the 'PdfRectangle' creator public


src/UglyToad.PdfPig/Content/Line.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Text;
+using UglyToad.PdfPig.Geometry;
+using System.Linq;
+
+namespace UglyToad.PdfPig.Content
+{
+    /// <summary>
+    /// A line.
+    /// </summary>
+    public class Line
```
EliotJones: My worry with the name `Line` is that people might confuse it with the `PdfLine` / expect it to be a geometrical feature. Maybe we could use the name `TextLine` or something similar and make the class comment slightly more descriptive, e.g. a line of text in the document, or something.

src/UglyToad.PdfPig/Content/Word.cs
```diff
@@ -41,15 +51,18 @@ public Word(IReadOnlyList<Letter> letters)
                 throw new ArgumentException("Empty letters provided.", nameof(letters));
             }
 
+            Letters = letters;
```
EliotJones: I'm surprised this wasn't here already, a good addition 👍

src/UglyToad.PdfPig/Content/Line.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            Text = string.Join(" ", words.Where(s => !string.IsNullOrWhiteSpace(s.Text)).Select(x => x.Text));
+
+            var minX = words.Min(x => x.BoundingBox.Left);
+            var minY = words.Min(x => x.BoundingBox.Bottom);
+            var maxX = words.Max(x => x.BoundingBox.Right);
+            var maxY = words.Max(x => x.BoundingBox.Top);
+            BoundingBox = new PdfRectangle(minX, minY, maxX, maxY);
+
+            TextDirection = words[0].TextDirection;
```
EliotJones: Do we need some additional guarantees here? I'm not sure if the user might expect text direction of all words in a line to be consistent, I'm trying to think of both possible use cases but it might be better to check `words.All(x => x.TextDirection == words[0].TextDirection)` or whatever the nicer way of expressing that is, I'll defer to you on this since you're working with the API.

src/UglyToad.PdfPig/Content/Line.cs
BobLd: Makes sense, will adjust that

src/UglyToad.PdfPig/Content/Line.cs
BobLd: Good point, will use `TextDirection = TextDirection.Unknown` if several directions are present.

src/UglyToad.PdfPig/Geometry/PdfLine.cs
```diff
@@ -0,0 +1,55 @@
...
+    public struct PdfLine
+    {
+        /// <summary>
+        /// Length of the line.
+        /// </summary>
+        public decimal Length
+        {
+            get
+            {
+                decimal l = (Point1.X - Point1.X) * (Point1.X - Point1.X) + (Point1.Y - Point1.Y) * (Point1.Y - Point1.Y);
```
EliotJones: Does this return the right value? Both the delta X and Y use point1 instead of point1 and point2?

src/UglyToad.PdfPig.Tests/PublicApiScannerTests.cs
```diff
@@ -45,6 +45,7 @@ public void OnlyExposedApiIsPublic()
                 "UglyToad.PdfPig.Content.Page",
                 "UglyToad.PdfPig.Content.PageSize",
                 "UglyToad.PdfPig.Content.Word",
+                "UglyToad.PdfPig.Content.Line",
```
EliotJones: It looks like this needs to be `TextLine` now to keep the API scanner test happy

PR #36, opened 2019-06-16
Document Layout Analysis Tools

BobLd:
Example of use of the `RecursiveXYCut`:

           using (PdfDocument document = PdfDocument.Open(path, options))
            {
                for (var i = 0; i < document.NumberOfPages; i++)
                {
                    var words = page.GetWords();

                    var cuts = RecursiveXYCut.GetBlocks(words , page.Width / 5m,
                        k => Math.Round(k.Mode(), 3), k => Math.Round(k.Mode() * 1.5m, 3));

                    var leefs = cuts.GetLeefs();

                    foreach (var leef in leefs)
                    {
                        var lines = leef.GetLines();

                        foreach (var line in lines)
                        {
                            Console.WriteLine("'" + line.Text + "'");
                        }
                    }
                }
            }


EliotJones:
Thanks very much, this looks great and as discussed over on the other issue should prove helpful for others. This week is looking a bit busy for me but I'll try to make time for this as soon as possible.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// The weighted Euclidean distance.
+        /// </summary>
+        /// <param name="point1">The first point.</param>
+        /// <param name="point2">The second point.</param>
+        /// <param name="wX">The weight of the X coordinates. Default is 1.</param>
+        /// <param name="wY">The weight of the Y coordinates. Default is 1.</param>
+        /// <returns></returns>
+        public static double WghtdEuclidean(PdfPoint point1, PdfPoint point2, double wX = 1.0, double wY = 1.0)
```
EliotJones: I think `WeightedEuclidean` would be a better name here, the abbreviation doesn't save enough typing to justify the extra reading effort.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        /// <summary>
+        /// The weighted Euclidean distance.
+        /// </summary>
+        /// <param name="point1">The first point.</param>
+        /// <param name="point2">The second point.</param>
+        /// <param name="wX">The weight of the X coordinates. Default is 1.</param>
+        /// <param name="wY">The weight of the Y coordinates. Default is 1.</param>
+        /// <returns></returns>
```
EliotJones: I'd prefer removing the empty doc-comment tags such as `<returns>` or empty `<param>` tags than leaving them in.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Find the nearest point.
+        /// </summary>
+        /// <param name="pdfPoint">The reference point, for which to find the nearest neighbour.</param>
+        /// <param name="points">The list of neighbours candidates.</param>
+        /// <param name="measure">The distance measure to use.</param>
+        /// <param name="dist">The distance between reference point, and its nearest neighbour</param>
+        /// <returns></returns>
+        public static PdfPoint FindNearest(this PdfPoint pdfPoint, PdfPoint[] points,
```
EliotJones: I think here it would be better to use `IReadOnlyList<PdfPoint>` over `PdfPoint[]` since it allows the consumer to use a `List<PdfPoint>` or some custom type while providing the guarantee we won't mutate their collection inside our method. There's a slight performance penalty to doing that (which I'm about to contradict myself on) but I think in this case the benefits to the publicly visible API are worth it.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+            Func<PdfPoint, PdfPoint, double> measure, out double dist)
+        {
+            double d = points.Min(k => measure(k, pdfPoint));
```
EliotJones: Might be worth a few precondition checks here with detailed exceptions, one to check `points` and `measure` (maybe `distanceCalculator` is a better name?) for null and one to check `points` is not empty (in which case the call to min fails but the exception from the Linq could always be more detailed :'( ).

EliotJones: On performance I think I'd prefer the following:

```suggestion
            var minDistance = double.MaxValue;
            Point closestPoint = default;
            for (var i = 0; i < points.Length; i++)
            {
                var distance = measure(points[i], pdfPoint);
                if (distance < minDistance) 
                {
                   minDistance = distance;
                   closestPoint = points[i];
                }
            }
            
            return closestPoint;
```

Where `points.Length` is `points.Count` for the `IReadOnlyList<PdfPoint>` case. This prevents needing to iterate over the collection twice with a call to `measure` on each iteration.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// Find the nearest point.
+        /// </summary>
+        /// <param name="pdfPoint">The reference point, for which to find the nearest neighbour.</param>
+        /// <param name="points">The list of neighbours candidates.</param>
+        /// <param name="measure">The distance measure to use.</param>
+        /// <param name="dist">The distance between reference point, and its nearest neighbour</param>
+        /// <returns></returns>
+        public static PdfPoint FindNearest(this PdfPoint pdfPoint, PdfPoint[] points,
+            Func<PdfPoint, PdfPoint, double> measure, out double dist)
```
EliotJones: On the abbreviation here I think `distance` is probably worth the extra characters again.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Find the index of the nearest point.
+        /// </summary>
+        /// <param name="pdfPoint">The reference point, for which to find the nearest neighbour.</param>
+        /// <param name="points">The list of neighbours candidates.</param>
+        /// <param name="measure">The distance measure to use.</param>
+        /// <param name="dist">The distance between reference point, and its nearest neighbour</param>
+        /// <returns></returns>
+        public static int FindIndexNearest(this PdfPoint pdfPoint, PdfPoint[] points,
```
EliotJones: The same comments apply here from the `FindNearest` method I think.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/MathExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Useful math extensions.
+    /// </summary>
+    public static class MathExtensions
+    {
+        /// <summary>
+        /// Computes the mode of a sequence of float values.
+        /// </summary>
+        /// <param name="array"></param>
```
EliotJones: Just some empty doc-comments to purge in here, the `Mode` enumerates twice but I'm ok with it for the time being since it's a bit of pain to unroll.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/NNWordExtractor.cs
```diff
@@ -0,0 +1,200 @@
...
+using UglyToad.PdfPig.Geometry;
+using UglyToad.PdfPig.Util;
+
+namespace UglyToad.PdfPig.DocumentLayoutAnalysis
+{
+    /// <summary>
+    /// Nearest Neighbour Word Extractor, using the <see cref="Distances.Manhattan"/> distance.
+    /// This implementation leverages bounding boxes.
+    /// </summary>
+    public class NNWordExtractor : IWordExtractor
```
EliotJones: On naming here (I'm annoying sorry!) I think the full name is going to help public api discovery:

```suggestion
    public class NearestNeighbourWordExtractor : IWordExtractor
```

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -0,0 +1,411 @@
...
+            {
+                if ((words[i].BoundingBox.Left >= currentProj[0] && words[i].BoundingBox.Left <= currentProj[1])
+                    || (words[i].BoundingBox.Right >= currentProj[0] && words[i].BoundingBox.Right <= currentProj[1]))
+                {
+                    // it is overlapping 
+                    if (words[i].BoundingBox.Left >= currentProj[0]
+                        && words[i].BoundingBox.Left <= currentProj[1]
+                        && words[i].BoundingBox.Right > currentProj[1])
+                    {
+                        // |____|
```
EliotJones: These little diagrams are a nice touch 👍

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -0,0 +1,411 @@
...
+
+    /// <summary>
+    /// A Node used in the <see cref="RecursiveXYCut"/> algorithm.
+    /// </summary>
+    public class XYNode
+    {
+        /// <summary>
+        /// Returns true if this node is a leef, false otherwise.
+        /// </summary>
+        public virtual bool IsLeef => false;
```
EliotJones: I think the naming should be `Leaf` here and in other places unless leef has a special meaning in this context?

src/UglyToad.PdfPig/DocumentLayoutAnalysis/Distances.cs
EliotJones: Actually having thought about this it's probably even more convenient to make the argument:

```suggestion
        public static PdfPoint FindNearest(this PdfPoint pdfPoint, IEnumerable<PdfPoint> points,
```

Which would make the suggested algorithm below use a `foreach` loop (throwing if there's not at least one item in the enumerable).

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
BobLd: Hehe makes sense, it's a typo indeed

BobLd:
A quick example of the result, with the proposed reading order indicated in red (to be improved):
A Study In Scarlet_small

src/UglyToad.PdfPig/DocumentLayoutAnalysis/NearestNeighbourWordExtractor .cs
```diff
@@ -0,0 +1,199 @@
...
+
+            List<Word> words = new List<Word>();
+            for (int a = 0; a < groupedIndexes.Count(); a++)
+            {
+                List<Letter> groupedLetters = new List<Letter>();
+                foreach (int s in groupedIndexes[a])
+                {
+                    groupedLetters.Add(letters[s]);
+                }
+                words.Add(new Word(groupedLetters));
```
EliotJones: Is it possible for letters to be out of order at this point? I think Linq operations don't guarantee order so even though I think in normal flow the letters will be in order it might be worth an explicit orderby here?

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -0,0 +1,408 @@
...
+    {
+        /// <summary>
+        /// Get the blocks.
+        /// </summary>
+        /// <param name="pageWords">The words in the page.</param>
+        /// <param name="minimumWidth">The minimum width for a block.</param>
+        /// <param name="dominantFontWidth">The dominant font width.</param>
+        /// <param name="dominantFontHeight">The dominant font height.</param>
+        /// <returns></returns>
+        public static XYNode GetBlocks(IEnumerable<Word> pageWords, decimal minimumWidth,
```
EliotJones: Is it worth an additional overload which takes just a `Page` and uses some preconfigured but overloadable defaults to help people fall into the pit of success?

e.g. 

```
public static XYNode GetBlocks(Page page, decimal? minimumWidth = null)
{
   return GetBlocks(page.GetWords(), minimumWidth.HasValue ? minimumWidth.Value : page.Width / 5, k => k.Max(), ...);
}```

Or something like that so that users can just 'call it' without too much knowledge and begin experimenting from there.

Comment (BobLd) [reply src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs]: yes, good idea but I think it's better to do the following:
`public static XYNode GetBlocks(IEnumerable<Word> pageWords, decimal minimumWidth = 0) { ... }`

File: src/UglyToad.PdfPig/DocumentLayoutAnalysis/NearestNeighbourWordExtractor .cs (hunk 11 above)
Comment (BobLd) [reply src/UglyToad.PdfPig/DocumentLayoutAnalysis/NearestNeighbourWordExtractor .cs]: Not sure if possible but will check that

File: src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs [hunk 13]
```diff
@@ -0,0 +1,408 @@
...
+    {
+        /// <summary>
+        /// Returns true if this node is a leaf, false otherwise.
+        /// </summary>
+        public override bool IsLeaf => true;
+
+        /// <summary>
+        /// The words in the leaf.
+        /// </summary>
+        public Word[] Words { get; set; }
```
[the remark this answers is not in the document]
BobLd: Agreed on the `IReadOnlyList`, need to check for the `set`

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// <returns></returns>
+        public static XYNode GetBlocks(IEnumerable<Word> pageWords, decimal minimumWidth,
+            Func<IEnumerable<decimal>, decimal> dominantFontWidthFunc,
+            Func<IEnumerable<decimal>, decimal> dominantFontHeightFunc)
+        {
+            var root = new XYLeaf(pageWords);
+            return VerticalCut(root, minimumWidth, dominantFontWidthFunc, dominantFontHeightFunc);
+        }
+
+        private static XYNode VerticalCut(XYLeaf leaf, decimal minimumWidth,
```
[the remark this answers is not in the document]
BobLd: **Abstract from the research paper** *(steps 1 and 2 are not necessary, they assume the pdf is an image)*:
1. Do preprocessing (eg. noise removal, skew correction, etc.).
2. Apply a connected component labeling algorithm.
3. Obtain the bounding boxes of connected components.
4. Create a root node.
5. Calculate the horizontal and vertical projection profiles within the region of interest.
6. Do divisions at large gaps in the projection profiles whose widths exceed a certain threshold value V. Whenever divisions are made, create a new child node. At each recursion level, horizontal and vertical divisions alternate.
7. Do Step 5-8 recursively until no further divisions are possible.

The threshold value V is determined by the dominant font size.

Illustration of the projection profiles:
projection profile
Figure 1: Horizontal Projection Profile, which is obtained by accumulating the bounding boxes onto the vertical line. The projection profile is a histogram which shows frequencies of projected bounding boxes.

Source: http://haralick.org/conferences/71280952.pdf

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -0,0 +1,408 @@
...
+        {
+            var groupedWords = Words.GroupBy(x => x.BoundingBox.Bottom).ToDictionary(x => x.Key, x => x.ToList());
+            return groupedWords.OrderByDescending(x => x.Key).Select(ws => new TextLine(ws.Value)).ToArray();
+        }
+
+        /// <summary>
+        /// Create a new <see cref="XYLeaf"/>.
+        /// </summary>
+        /// <param name="words">The words contained in the leaf.</param>
+        public XYLeaf(params Word[] words) : this(words == null ? null : words.ToList())
```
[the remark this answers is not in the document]
BobLd: I think it's better to leave it like this so that it does not throw here. The check will be added in `public XYLeaf(IEnumerable<Word> words) : base(null)`

Issue #32, opened 2019-06-04
Issues reading unicode document properties?


Numpsy:
Hi,

I had a try with loading some simple test documents into PdfPig 0.0.6, and noticed that unicode document properties don't seem to be handled correctly.

e.g., if I open the attached [minimal.pdf](https://github.com/UglyToad/PdfPig/files/3253137/minimal.pdf) in Acrobat reader it displays:

![image](https://user-images.githubusercontent.com/1178570/58890904-6adbd200-86e3-11e9-88a2-03bca07501f2.png)

but in PdfDocument.Information, I get:

![image](https://user-images.githubusercontent.com/1178570/58891209-053c1580-86e4-11e9-835a-4339b7a32503.png)

Would this be expected to work?

Thanks.

EliotJones:
Thanks for raising this and providing a replication. 

I've had a look and it seems PDFs can support UTF-16 Big and Little Endian strings in its content, I've added what I believe should fix the behaviour here though it seems like the resulting string is 'weird' when it comes to comparing string equality with normal C# strings.

Numpsy:
Thanks, that looks better now.

I have just accidentally discovered a variation of the issue when trying to create a new test pdf with LibreOffice though - the document properties come out as HexTokens with a BOM attached:

![HexToken](https://user-images.githubusercontent.com/1178570/59858912-b8a73a00-9373-11e9-837b-9cc81b0746da.png)

Sample file: 
[PropsAddedWithLibreOffice.pdf](https://github.com/UglyToad/PdfPig/files/3310951/PropsAddedWithLibreOffice.pdf)

Numpsy:
> 
> though it seems like the resulting string is 'weird' when it comes to comparing string equality with normal C# strings.

The [MS Docs](https://docs.microsoft.com/en-us/dotnet/api/system.text.unicodeencoding.getstring?view=netframework-4.8) suggests that the strings that come back from GetString have a BOM left at the start, and suggests Trimming the string to remove them?

EliotJones:
Ok, I've made it remove the BOM by default now, hopefully this won't break something else! 😆

EliotJones:
Does this seem to be working for the documents you're dealing with after the fix and if so are you happy to close it? Just finding old issues to close where I can.

Numpsy:
Looks good now

PR #40, opened 2019-06-23
Add UTF-16 parsing support to HexToken

Numpsy:
This is an attempt at handling the UTF-16 HexToken issue mentioned in #32, by changing the HexToken constructor to look for the UTF16BE BOM at the start of the data and decoding it.

It also adds a couple of unit tests. The change is enough to make those pass, but i'm not sure if any other changes are needed (I know very little about the structure of PDFs).

### DIFF
--- a/src/UglyToad.PdfPig/Tokens/HexToken.cs
+++ b/src/UglyToad.PdfPig/Tokens/HexToken.cs
@@ -62,7 +62,6 @@ namespace UglyToad.PdfPig.Tokens
             }
 
             var bytes = new List<byte>();
-            var builder = new StringBuilder();
 
             for (var i = 0; i < characters.Count; i += 2)
             {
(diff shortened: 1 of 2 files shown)


src/UglyToad.PdfPig/Tokens/HexToken.cs
```diff
@@ -79,15 +78,29 @@ public HexToken([NotNull] IReadOnlyList<char> characters)
...
                 var b = Convert(high, low);
                 bytes.Add(b);
+            }
+
+            // Handle UTF-16BE format strings.
+            if (bytes.Count >= 2 && bytes[0] == 0xFE && bytes[1] == 0xFF)
+            {
+                Data = Encoding.BigEndianUnicode.GetString(bytes.ToArray(), 2, bytes.Count - 2);
+            }
+            else
```
EliotJones: There's another case where the string uses LE Unicode starting with (hex) `FFFE` which it's probably worth adding as an extra `else if` here. I haven't seen it in the wild yet but I think it's worth covering:

```               
else if (bytes.Count >= 2 && bytes[0] == 0xFF && bytes[1] == 0xFE)
                {
                    Data = Encoding.Unicode.GetString(bytes.ToArray(), 2, bytes.Count - 2);
                }```

EliotJones:
Thanks for taking the time to do this (and for adding tests 😄), I'm a bit worried about the potential for a hex token appearing with the `FEFF` or `FFFE` sequence in a page content stream (the sequence of bytes indicating the pseudo-postscript operations used to build a page) rather than the main PDF content but it's not something I've seen and I don't think it's worth worrying about for the time being.

If you get a chance to add the extra `else if` case mentioned in the other comment it should be good to merge, let me know if you won't get time and I'll merge anyway.

src/UglyToad.PdfPig/Tokens/HexToken.cs
Numpsy: I'd been attempting to read the pdf spec and wasn't sure if the LE case was needed here (the 'text string type'/ section 7.9.2.2 of the spec I was looking at only mentioned the BE case).
I can add that case later if you want, though I don't have any 'real' test data for it.

EliotJones: Ah, fair enough, yeah it's not the clearest so I'll merge this fix in then in future it's easy to fix if the LE case does turn out to be needed.

Issue #38, opened 2019-06-19
Support for custom document properties in PDFs?


Numpsy:
Hi,

Can I ask if there are any plans and/or possibility of supporting custom document properties in PDF files as well as the 'known' ones (Author and Keywords and such)?

I haven't used PDFBox, but the documentation for PDDocumentInformation does seem to have functions to access custom properties.

EliotJones:
That sounds like a good addition to the API, it would be fairly easy to add by providing access to the underlying `DictionaryToken` in the `DocumentInformation` object. I'll look into adding it when I get time.

EliotJones:
This is in the latest master code now, it might be worth testing it out for your documents if you're able, I don't have any documents with custom info at the moment to test against. It should make it into the 0.0.7 release for which the timeline is "hopefully soonish".

Numpsy:
Thanks, i'll give it a test tommorow.

Numpsy:
Ok, I started adding a unit test @ https://github.com/Numpsy/PdfPig/commit/db0f1c7fb0d0bc27d5aeccd91aefde8cef71eb2b, and i'm not sure if I've got another issue there.

The 'custom-properties.pdf' file attached to that commit contains multiple sets of document information, which have a few differences in the property values.
If I open the file in Firefox, it shows the Keywords property as being 'MoreKeywords' (which is what I was expecting), but when loading in PdfPig the value in Document.Information is 'TestKeywords', which is an old/previous value.

Possibly something else going wrong somewhere?

EliotJones:
Thanks for the further bug report, I've pushed some changes which seem to work for this document (and not break the tests!). Let me know if you run into any more issues please, I think the code in that area needs a bit of a rewrite/tidy-up.

Numpsy:
Ok, that seems to work now, thanks.

Would you be interested in a PR for https://github.com/Numpsy/PdfPig/commit/10dcbd0cc4d836dd93bb18634af918ae7411992b to add a unit test?

EliotJones:
Yes, thank you, that would be good to cover with a test.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

EliotJones:
Just doing some tidying up, are you happy to close this issue now?

Numpsy:
Yes, it looks good now, i'll close the issue.

PR #43, opened 2019-07-06
Add a unit test for reading document information

Numpsy:
Add a unit test for reading custom properties from the document information (#38)


EliotJones:
Thanks, this looks good and it should help refactoring this area in future when I/anyone else gets round to it. Thanks again for all your bug reports, they've been extremely useful in making the library more stable for everyone.

I'm just going to wait to see if Appveyor will successfully build this, I don't think it's your changes that are breaking the build, it's just that at weekends the build agents get busy and the builds start timing out (even though the timeout is set to 1hr 30m!). If it doesn't build after a couple of tries I'll merge it anyway.

Numpsy:
I wondered what the situation was with the builds timing out (I mean, given that they don't seem to take much time at all when I run them locally).

EliotJones:
Yeah I'm not really sure what might be causing these timeouts. For now I assume they are due to disk access being very slow on the build machine or something. I imagine the free tier Appveyor build agents are the lowest spec possible.

Hopefully there'll be a free alternative to Appveyor for Windows builds soon, maybe Github will add something.

Since Appveyor has clearly given up today I'll merge this now and check it runs locally.

Numpsy:
> Hopefully there'll be a free alternative to Appveyor for Windows builds soon, maybe Github will add something.
> 

Azure Pipelines has some free support for open source projects, but I couldn't say how performance compares with Appveyor.

PR #46, opened 2019-07-16
Adding Paths collection

vadik299:
Hi, I added Paths collection to Page to be able to get the list of lines/curves on the page. Also added matrix transform to the path operators.

### DIFF
--- a/src/UglyToad.PdfPig/Content/Page.cs
+++ b/src/UglyToad.PdfPig/Content/Page.cs
@@ -45,7 +45,7 @@
         /// <summary>
         /// The set of <see cref="PdfPath"/>s drawn by the PDF content.
         /// </summary>
-        public List<PdfPath> Paths => Content?.Paths ?? new List<PdfPath>();
+        public IReadOnlyList<PdfPath> Paths => Content?.Paths ?? new List<PdfPath>();
 
         /// <summary>
         /// The full text of all characters on the page in the order they are presented in the PDF content.
--- a/src/UglyToad.PdfPig/Content/Page.cs
+++ b/src/UglyToad.PdfPig/Content/Page.cs
@@ -9,6 +9,7 @@
     using Util;
     using Util.JetBrains.Annotations;
     using XObjects;
+    using UglyToad.PdfPig.Geometry;
 
     /// <summary>
     /// Contains the content and provides access to methods of a single page in the <see cref="PdfDocument"/>.
@@ -40,6 +41,11 @@
         /// The set of <see cref="Letter"/>s drawn by the PDF content.
         /// </summary>
         public IReadOnlyList<Letter> Letters => Content?.Letters ?? new Letter[0];
+
+        /// <summary>
+        /// The set of <see cref="PdfPath"/>s drawn by the PDF content.
+        /// </summary>
+        public List<PdfPath> Paths => Content?.Paths ?? new List<PdfPath>();
 
         /// <summary>
         /// The full text of all characters on the page in the order they are presented in the PDF content.
--- a/src/UglyToad.PdfPig/Content/PageContent.cs
+++ b/src/UglyToad.PdfPig/Content/PageContent.cs
@@ -24,7 +24,7 @@
         internal IReadOnlyList<IGraphicsStateOperation> GraphicsStateOperations { get; }
 
         public IReadOnlyList<Letter> Letters { get; }
-        public List<PdfPath> Paths { get; }
+        public IReadOnlyList<PdfPath> Paths { get; }
 
         internal PageContent(IReadOnlyList<IGraphicsStateOperation> graphicsStateOperations, IReadOnlyList<Letter> letters, List<PdfPath> paths,
             IReadOnlyDictionary<XObjectType, List<XObjectContentRecord>> xObjects,
--- a/src/UglyToad.PdfPig/Content/PageContent.cs
+++ b/src/UglyToad.PdfPig/Content/PageContent.cs
@@ -5,6 +5,7 @@
     using Graphics.Operations;
     using Tokenization.Scanner;
     using XObjects;
+    using UglyToad.PdfPig.Geometry;
 
     /// <summary>
     /// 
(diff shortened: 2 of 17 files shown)


EliotJones:
Thanks very much for putting this together, I'll try to get round to reviewing it as soon as I can.

src/UglyToad.PdfPig/Content/Page.cs
```diff
@@ -41,6 +42,11 @@ public class Page
         /// </summary>
         public IReadOnlyList<Letter> Letters => Content?.Letters ?? new Letter[0];
 
+        /// <summary>
+        /// The set of <see cref="PdfPath"/>s drawn by the PDF content.
+        /// </summary>
+        public List<PdfPath> Paths => Content?.Paths ?? new List<PdfPath>();
```
EliotJones: As a general rule I avoid exposing the concrete types `List<T>` and `T[]` on the public API. This is because the consumer can mutate the content of those types and potentially break the code using those values.

Most of the time `IReadOnlyList<T>`  provides strict enough guarantees for a public API and also means we can change the internal code to use a list, array, immutablearray, etc.

This would work like this:
```suggestion
        public IReadOnlyList<PdfPath> Paths => Content?.Paths ?? new List<PdfPath>();
```

src/UglyToad.PdfPig/Content/PageContent.cs
```diff
@@ -23,15 +24,17 @@ internal class PageContent
         internal IReadOnlyList<IGraphicsStateOperation> GraphicsStateOperations { get; }
 
         public IReadOnlyList<Letter> Letters { get; }
+        public List<PdfPath> Paths { get; }
```
EliotJones: Same here, I'd change this to:
```suggestion
        public IReadOnlyList<PdfPath> Paths { get; }
```

src/UglyToad.PdfPig/Graphics/Operations/PathConstruction/AppendRectangle.cs
```diff
@@ -52,7 +52,7 @@ public void Run(IOperationContext operationContext)
         {
             operationContext.BeginSubpath();
             operationContext.CurrentPath.Rectangle(LowerLeft.X, LowerLeft.Y, Width, Height);
-            operationContext.CurrentPath.ClosePath();
+            //operationContext.CurrentPath.ClosePath();
```
EliotJones: If possible delete this commented out code please. I'm just going to check why the tests are failing because I think it might be related.

src/UglyToad.PdfPig/Graphics/Operations/PathConstruction/BeginNewSubpath.cs
```diff
@@ -37,6 +37,7 @@ public void Run(IOperationContext operationContext)
         {
             operationContext.BeginSubpath();
             operationContext.CurrentPosition = Point;
+            operationContext.CurrentPath.LineTo(Point.X, Point.Y);
```
EliotJones: This doesn't quite seem like the right behavior here, the spec says:

> Begin a new subpath by moving the current point to coordinates (x, y), omitting any connecting line segment. If the previous path construction operator in the current path was also m, the new m overrides it; no vestige of the previous m operation remains in the path.

It might be we need to support some other operation on the `CurrentPath` object.

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -5,27 +5,29 @@ namespace UglyToad.PdfPig.Geometry
...
 
     /// <summary>
     /// A path in a PDF document, used by glyphs and page content.
     /// </summary>
     public class PdfPath
     {
-        private readonly List<IPathCommand> commands = new List<IPathCommand>();
+        public readonly List<IPathCommand> Commands = new List<IPathCommand>();
         private PdfPoint? currentPosition;
+        internal TransformationMatrix CurrentTransformationMatrix;
```
EliotJones: This is a good addition 👍 

I'd make this a constructor parameter rather than an `internal` settable field. As far as possible in the code I try to avoid publically mutable fields/properties. It makes the code easier to reason about if it can't be changed from elsewhere.

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -5,27 +5,29 @@ namespace UglyToad.PdfPig.Geometry
...
     using System.Text;
+    using UglyToad.PdfPig.Core;
 
     /// <summary>
     /// A path in a PDF document, used by glyphs and page content.
     /// </summary>
     public class PdfPath
     {
-        private readonly List<IPathCommand> commands = new List<IPathCommand>();
+        public readonly List<IPathCommand> Commands = new List<IPathCommand>();
```
EliotJones: I'd keep this `private readonly` and have a property that exposes a public immutable version:
```public IReadOnlyList<IPathCommand> Commands => commands;```

File: src/UglyToad.PdfPig/Content/Page.cs (hunk 1 above)
Comment (EliotJones) [reply src/UglyToad.PdfPig/Content/Page.cs]: It might be worth exposing this through the `ExperimentalAccess` member of `Page` for now. I'm still not sure what the right API for these is (for example I was planning on using lines to support underlined text detection as well as incorporating color).

File: src/UglyToad.PdfPig/Geometry/PdfPath.cs [hunk 7]
```diff
@@ -5,27 +5,29 @@ namespace UglyToad.PdfPig.Geometry
...
         private PdfPoint? currentPosition;
+        internal TransformationMatrix CurrentTransformationMatrix;
 
         internal void MoveTo(decimal x, decimal y)
         {
-            currentPosition = new PdfPoint(x, y);
-            commands.Add(new Move(currentPosition.Value));
+            currentPosition = CurrentTransformationMatrix.Transform(new PdfPoint(x, y));
```
EliotJones: So it looks like the test failures are due to `CurrentTransformationMatrix` not being set prior to these calls when it's used by `CharString` operations in font handling text (Type 1 fonts are formed of path operations). I think it would be better to move these transformations out into the `OperationContext` so that `x` and `y` have already been transformed prior to passing into the `MoveTo`, `LineTo`, etc, methods.
To make this easier you could add a `TransformationMatrix CurrentTransformationMatrix { get; }` property to the `IOperationContext` interface.

vadik299:
Thank you for the comments! I will review the code and resolve the issues. Sorry, been a bit busy lately :-(

src/UglyToad.PdfPig/Graphics/Operations/PathConstruction/BeginNewSubpath.cs
vadik299: I think it's fine. There's logic in LineTo that actually does MoveTo when current position is not set yet.

PR #50, opened 2019-08-06
 Document Layout Analysis - Text edges extractor

BobLd:
Text edges are where words have either there BoundingBox's left, right or mid coordinates aligned on the same vertical line. Useful to detect tables, justified text, lists, etc.

Example uf use:

            using (PdfDocument document = PdfDocument.Open(path))
            {
                for (var i = 0; i < document.NumberOfPages; i++)
                {
                    var words = page.GetWords();
                    var edges = TextEdgesExtractor.GetEdges(words);

                    foreach (var edge in edges["left"])
                    {
                        // Do something with left edges
                    }

                    foreach (var edge in edges["mid"])
                    {
                        // Do something with mid edges
                    }

                    foreach (var edge in edges["right"])
                    {
                        // Do something with right edges
                    }
                }
            }

Example of result:
Random 2 Columns Lists
Blue: Left edge
Green: Mid edge (see dotted list)
Pink: Right edge

### DIFF
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
@@ -9,8 +9,8 @@ using UglyToad.PdfPig.Geometry;
 namespace UglyToad.PdfPig.DocumentLayoutAnalysis
 {
     /// <summary>
-    /// Text edges extractor. Text edges are where words have either there BoundingBox's left, right or mid coordinate aligned on the same vertical line.
-    /// <para>Useful to detect tables, justified text, lists, etc.</para>
+    /// Text edges extractor. Text edges are where words have either there BoundingBox's left, right or mid coordinates aligned on the same vertical line.
+    /// <para>Useful to detect text columns, tables, justified text, lists, etc.</para>
     /// </summary>
     public class TextEdgesExtractor
     {
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
@@ -80,4 +80,25 @@ namespace UglyToad.PdfPig.DocumentLayoutAnalysis
             return cleanEdges.SelectMany(x => x.Value.Select(y => new PdfLine(x.Key, y.Min(w => w.BoundingBox.Bottom), x.Key, y.Max(w => w.BoundingBox.Top)))).ToList();
         }
     }
+
+    /// <summary>
+    /// The type of edge.
+    /// </summary>
+    public enum EdgeType
+    {
+        /// <summary>
+        /// Text edges where words have their BoundingBox's left coordinate aligned on the same vertical line.
+        /// </summary>
+        Left = 0,
+
+        /// <summary>
+        /// Text edges where words have their BoundingBox's mid coordinate aligned on the same vertical line.
+        /// </summary>
+        Mid = 1,
+
+        /// <summary>
+        /// Text edges where words have their BoundingBox's right coordinate aligned on the same vertical line.
+        /// </summary>
+        Right = 2
+    }
 }
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
@@ -9,7 +9,7 @@ using UglyToad.PdfPig.Geometry;
 namespace UglyToad.PdfPig.DocumentLayoutAnalysis
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+using UglyToad.PdfPig.Content;
+using UglyToad.PdfPig.Geometry;
+
+namespace UglyToad.PdfPig.DocumentLayoutAnalysis
+{
+    /// <summary>
+    /// Text edges extractor. Text edges are where words have either there BoundingBox's left, right or mid coordinates aligned on the same vertical line.
+    /// <para>Useful to detect text columns, tables, justified text, lists, etc.</para>
+    /// </summary>
+    public class TextEdgesExtractor
```
EliotJones: Probably worth making this a `public static class` so that consumers don't try and new it up and get confused where the methods are, just reduces the discovery surface area of the public API.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using UglyToad.PdfPig.Content;
+using UglyToad.PdfPig.Geometry;
+
+namespace UglyToad.PdfPig.DocumentLayoutAnalysis
+{
+    /// <summary>
+    /// Text edges extractor. Text edges are where words have either there BoundingBox's left, right or mid coordinates aligned on the same vertical line.
```
EliotJones: Just a typo on `there` <-> `their` here

src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            var cleanWords = pageWords.Where(x => !string.IsNullOrWhiteSpace(x.Text.Trim()));
+
+            ConcurrentDictionary<string, List<PdfLine>> dictionary = new ConcurrentDictionary<string, List<PdfLine>>();
+
+            Parallel.ForEach(edgesFuncs, f =>
+            {
+                dictionary.TryAdd(f.Item1, GetVerticalEdges(cleanWords, f.Item2, minimumElements));
+            });
+            return dictionary.ToDictionary(x => x.Key, x => x.Value);
```
EliotJones: One thing it is perhaps worth doing here is using enum instead of string values for the key in the returned and internal dictionary, so an enum inside this class, `TextEdgesExtractor.EdgeType` like:

```
public enum EdgeType
{
   Left = 0,
   Mid = 1,
   Right = 2
}
```

This avoids people having to know up-front the correct case and terminology for the keys in this dictionary (e.g. is it `Mid`, `mid`, `MID`, `Middle`, ...).

src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+            Tuple.Create<string, Func<PdfRectangle, decimal>>("mid", x => Math.Round(x.Left + x.Width / 2, 0)),     // use BoundingBox's mid coordinate
+            Tuple.Create<string, Func<PdfRectangle, decimal>>("right",  x => Math.Round(x.Right, 0))                // use BoundingBox's right coordinate
+        };
+
+        /// <summary>
+        /// Get the text edges.
+        /// </summary>
+        /// <param name="pageWords">The words in the page.</param>
+        /// <param name="minimumElements">The minimum number of elements to define a text edge.</param>
+        public static Dictionary<string, List<PdfLine>> GetEdges(IEnumerable<Word> pageWords, int minimumElements = 4)
```
EliotJones: I think this is probably a no because `uint` and the unsigned types are kind of ugly but is it worth using a `uint` here since a negative minimumElements is probably wrong. I think the code would handle it fine so I'm not fussed.

EliotJones: Because I'm an immutable nerd I'd prefer the return type here to be an `IReadOnlyDictionary<TKey, TValue>`, up to you though.

src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
BobLd: yes much better

src/UglyToad.PdfPig/DocumentLayoutAnalysis/TextEdgesExtractor.cs
BobLd: I've added the following check:
 `if (minimumElements < 0)
{
      throw new ArgumentException("TextEdgesExtractor.GetEdges(): The minimum number of elements should be positive.", "minimumElements");
}`

BobLd:
Should be good to go!

PR #51, opened 2019-08-06
Update README.md

BenyErnest:
I think you need to pass the byte array to the File.WriteAllBytes method.


EliotJones:
Thanks, good spot, for some reason I always forget the second parameter with that method 😆

BenyErnest:
Glad to help.

PR #52, opened 2019-08-08
Change StreamInputBytes.Seek to reset isAtEnd to false

Numpsy:
When trying to do some tests with 0.0.7 I noticed that some files were failing to open when using the variant of PdfDocument.Open that takes a stream, but were working when using the variant that takes a file path.

Debugging through the code, I think that might be because StreamInputBytes doesn't reset isAtEnd when the stream is seeked away from the end, so once the end has been reached isAtEnd gets left at true permanently.


EliotJones:
Thanks, that's a big bug 😟

Numpsy:
> 
> Thanks, that's a big bug 😟

I'd been hoping to do some more tests a while back, but unfortunately only just got time to do it :-(

PR #54, opened 2019-08-09
Improving PdfPoint

BobLd:
Adding:
- `ToDouble`
- `Equals`
- `GetHashCode`

Not sure if `GetHashCode` is properly done, any feedback welcome.

### DIFF
--- a/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
+++ b/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
@@ -83,6 +83,39 @@
         }
 
         /// <summary>
+        /// Converts this <see cref="PdfPoint"/> into an array.
+        /// </summary>
+        /// <returns></returns>
+        public double[] ToDouble()
+        {
+            return new double[] { (double)this.X, (double)this.Y };
+        }
+
+        /// <summary>
+        /// Returns a value indicating whether this <see cref="PdfPoint"/> is equal to a specified <see cref="PdfPoint"/> .
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj is PdfPoint)
+            {
+                PdfPoint point = (PdfPoint)obj;
+                return point.X == this.X && point.Y == this.Y;
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Returns the hash code for this <see cref="PdfPoint"/>.
+        /// </summary>
+        /// <returns></returns>
+        public override int GetHashCode()
+        {
+            return (int)(this.X * 10_000 + 31 * this.Y * 10_000);
+        }
+
+        /// <summary>
         /// Get a string representation of this point.
         /// </summary>
         public override string ToString()
--- a/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
+++ b/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
@@ -83,14 +83,6 @@
         }
 
         /// <summary>
-        /// Converts this <see cref="PdfPoint"/> into an double array.
-        /// </summary>
-        public double[] ToDouble()
-        {
-            return new double[] { (double)this.X, (double)this.Y };
-        }
-
-        /// <summary>
         /// Returns a value indicating whether this <see cref="PdfPoint"/> is equal to a specified <see cref="PdfPoint"/> .
         /// </summary>
         /// <param name="obj"></param>
--- a/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
+++ b/src/UglyToad.PdfPig/Geometry/PdfPoint.cs
@@ -83,9 +83,8 @@
         }
 
         /// <summary>
-        /// Converts this <see cref="PdfPoint"/> into an array.
+        /// Converts this <see cref="PdfPoint"/> into an double array.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/UglyToad.PdfPig/Geometry/PdfPoint.cs
```diff
@@ -82,6 +82,39 @@ internal PdfVector ToVector()
...
+        }
+
+        /// <summary>
+        /// Returns a value indicating whether this <see cref="PdfPoint"/> is equal to a specified <see cref="PdfPoint"/> .
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public override bool Equals(object obj)
+        {
+            if (obj is PdfPoint)
```
EliotJones: There's a slightly terser syntax for this:

```suggestion
            if (obj is PdfPoint point)
```

Then you can remove line 103, but I don't mind, I just know Resharper is going to nag me 😆

src/UglyToad.PdfPig/Geometry/PdfPoint.cs
```diff
@@ -82,6 +82,39 @@ internal PdfVector ToVector()
             return new PdfVector(X, Y);
         }
 
+        /// <summary>
+        /// Converts this <see cref="PdfPoint"/> into an array.
+        /// </summary>
+        /// <returns></returns>
+        public double[] ToDouble()
```
EliotJones: I wonder if this would be better as a `ValueTuple` rather than an array. Since the array will always be known to have 2 members a tuple has a couple of benefits:
+ Prevents incorrect indexing by consumers
+ Avoids an allocation on the heap which needs garbage collection

This would look something like:
```suggestion
        /// <summary>Converts this <see cref="PdfPoint" to double values.<.summary>
        public (double x, double y) ToDouble()
```

I think PdfPig already includes ValueTuple as a dependency (it's a nuget package for earlier versions of the .NET Framework) so all consumers will be able to use it.

(secondary nitpick on the empty `<returns>` tags).

src/UglyToad.PdfPig/Geometry/PdfPoint.cs
```diff
@@ -82,6 +82,39 @@ internal PdfVector ToVector()
...
+            return false;
+        }
+
+        /// <summary>
+        /// Returns the hash code for this <see cref="PdfPoint"/>.
+        /// </summary>
+        /// <returns></returns>
+        public override int GetHashCode()
+        {
+            return (int)(this.X * 10_000 + 31 * this.Y * 10_000);
```
EliotJones: Overriding `GetHashCode` is one of those annoying things that we're told to do but correct approaches to it aren't really documented anywhere. There's a couple of discussions on it:

+ https://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode/263416#263416
+ https://blogs.msdn.microsoft.com/ericlippert/2011/02/28/guidelines-and-rules-for-gethashcode/

Initially I thought the approach here might be more vulnerable to collisions but now I'm not so sure, though I guess the only way to be sure is calculate the collision rate or something. I tend to treat hash codes as something like encryption in that they're much easier to use library functions for and easy to get wrong when written from scratch.

I generally rely on Resharper when doing this which generates something that closely matches Jon Skeet's answer. For reference this looks something like:

```
public override int GetHashCode()
{
    var hashCode = 1861411795;
    hashCode = hashCode * -1521134295 + X.GetHashCode();
    hashCode = hashCode * -1521134295 + Y.GetHashCode();
    return hashCode;
}
```

The other option is to let the .NET Framework do the heavy lifting by relying on the `ValueTuple` implementation:

```return (X, Y).GetHashCode();```

Which turns the point into a ValueTuple and uses the implementation from that which is here: https://github.com/dotnet/roslyn/blob/master/src/Compilers/Test/Resources/Core/NetFX/ValueTuple/ValueTuple.cs#L297. Now I'm not sure what that does but I'd hope it's one of the best possible implementations.

File: src/UglyToad.PdfPig/Geometry/PdfPoint.cs (hunk 1 above)
Comment (BobLd) [reply src/UglyToad.PdfPig/Geometry/PdfPoint.cs]: Very good, thx!

File: src/UglyToad.PdfPig/Geometry/PdfPoint.cs (hunk 3 above)
Comment (BobLd) [reply src/UglyToad.PdfPig/Geometry/PdfPoint.cs]: Hehe exactly. Let's .Net do the job then!

File: src/UglyToad.PdfPig/Geometry/PdfPoint.cs (hunk 2 above)
Comment (BobLd) [reply src/UglyToad.PdfPig/Geometry/PdfPoint.cs]: On this, I would really need a double array for computation purpose...

Comment (EliotJones) [reply src/UglyToad.PdfPig/Geometry/PdfPoint.cs]: Hmm, I'd be tempted to leave this off the public API for now. My gut instinct is that it'll confuse consumers of the library over the alternative of having a tuple. You could add:

````
public static class PdfPointExtensions
{
    public static double[] AsDoubleArray(this PdfPoint point) => new [] { (double)point.X, (double)point.Y };
}
```

Either to your consumer code or to PdfPig (or internally to PdfPig if it's for layout analysis code internally) but I'd avoid having it as a method directly on the public API of `PdfPoint` for now to reduce the amount of autocomplete options.

BobLd: okay, you are right. Let's keep it simple and leave it put for the moment

PR #56, opened 2019-08-10
Document Layout Analysis - IPageSegmenter, Docstrum

BobLd:
- Create a `TextBlock` class
- Create `IPageSegmenter`
- Add other useful distances: angle, etc.
- Update `RecursiveXYCut`:
   - With `IPageSegmenter` and `TextBlock`
   - Make `XYNode` and `XYLeaf` internal
- Optimise (faster) `NearestNeighbourWordExtractor` and isolate the clustering algorithms into `ClusteringAlgorithms.SimpleTransitiveClosure()` for use outside of this class
- Implement a Docstrum inspired page segmentation algorithm: `DocstrumBB`

Example of use of IPageSegmenter (e.g. RecursiveXYCut, DocstrumBB)

                 var words = page.GetWords(NearestNeighbourWordExtractor.Instance);
                 var blocks = DocstrumBB.Instance.GetBlocks(words);
                 foreach (var block in blocks)
                 {
                    // Do something with the block
                 }

Example of results (`Word` are in pink, `TextLine` in green, and `TextBlock` in blue):

`RecursiveXYCut`:
Random 2 Columns Lists XYCut

`DocstrumBB`:
Random 2 Columns Lists DocstrumBB
NB: When using `DocstrumBB` with justified text (as above), some words might not be linked as lines. Will fixe this in a second iteration.


src/UglyToad.PdfPig/DocumentLayoutAnalysis/DocstrumBB.cs
```diff
@@ -0,0 +1,212 @@
...
+                // Within-line distance
+                var pointWL = GetNearestPointData(pageWordsArr, word,
+                    bb => bb.BottomRight, bb => bb.BottomRight,
+                    bb => bb.BottomLeft, bb => bb.BottomLeft,
+                    wlAngleLB, wlAngleUB, Distances.Horizontal);
+                if (pointWL != null) withinLineDistList.Add(pointWL);
+
+                // Between-line distance
+                var pointBL = GetNearestPointData(pageWordsArr, word,
+                    bb => bb.BottomLeft, bb => bb.Centroid,
```
EliotJones: I wonder if this calculation of between line distance can be thrown off by letters with descenders/ascenders. I'm guessing because I'm not clever enough to understand the algorithm but would `letter.Location` work better since it represents the shared baseline of letters on a line: https://github.com/UglyToad/PdfPig/wiki/Letters

Though I guess that doesn't work for vertical/rotated text?

EliotJones: Oh this is words not letters, ignore me.

BobLd: Hi, event if it uses `Word` and not `Letter`, I'll have a look as descenders/ascenders might have an impact. Also, good point on Rotated and Vertical text, but needs more dev

BobLd:
I just need to make some modification on `DosctrumBB`, will update it shortly

BobLd:
NB: As I've put into the `DocstrumBB` class, the `FindIndexNearest()` can cause issues as there is a possibility that the closest point found is the point itself. To prevent this, we would need to find the k nearest neighbour, and remove the point from the found neighbours.

Maybe the best would be to use an existing implementation, like the following:
https://github.com/MathFerret1013/Supercluster.KDTree/tree/master/KDTree

BobLd:
Good to go for now, I keep your comments on descenders/ascenders and Rotated and Vertical text in mind.

EliotJones:
> To prevent this, we would need to find the k nearest neighbours, and remove the point from the found neighbours.

Would it be possible instead to have a flag which allows finding a point outside the max distance if the closest point is the item itself? I'll have a think about this. I'd avoid any 3rd party dependencies in the main project for ease of development/use/licensing reasons.

BobLd:
Agree with you, I was thinking to take the code and adapt it to our case, if license allows.

Unfortunately, the `FindIndexNearest()` cannot know that the point it found is the point itself in the current implementation. 

For example, if I use the `BottomRight` point of a current `Word` _A_ to find the closest `Word` using their `BottomLeft` point, the `BottomRight` of _A_ won't be in the list of candidates, and `FindIndexNearest()` can find _A_'s `BottomLeft` point if it's the closest. 

More importantly, if we use the same reference point to find the closest neighbour (e.g. using `BottomLeft` for both _A_ and the candidates), the `FindIndexNearest()` will **always** find the point itself...

Not sure if clear, but happy to discuss

PR #60, opened 2019-08-14
Improve ClusteringAlgorithms.GroupIndexes() and add Equals() to PdfLine

BobLd:


EliotJones:
Is this mainly performance improvements to groupindices or does it also change the behaviour, I wasn't able to follow it all the way?

BobLd:
Hi Eliot, only performance improvment, behaviour should be unchanged

PR #61, opened 2019-08-16
Adding TextSequence number to each letter to determine if letters belong to the same Tj operation

vadik299:
Hi, I believe it's useful to know whether the letters have been printed using same operation or not, so I added a simple sequence number to keep track of text printing operations.


EliotJones:
Thanks, this should be useful for users to determine word order, etc, if the producer includes it.

PR #68, opened 2019-10-04
Improve PdfPath

BobLd:
__PdfPath__
- Determine if path is closed 
- Determine if path is organised in a __clockwise__ or __counterclockwise__ manner (see accepted answer [here](https://stackoverflow.com/questions/1165647/how-to-determine-if-a-list-of-polygon-points-are-in-clockwise-order) for algo, and see [wiki](https://en.wikipedia.org/wiki/Shoelace_formula)). This is usefull to understand how shapes are filled (using `nonzero` fill rule). NB: it is possible for a closed path to be neither clockwise nor counterclockwise. Hence both `IsClockwise` and `IsCounterClockwise`.
- Add Centroid

__BezierCurve__
- Add `ToLines(int n)` to convert the curve to lines

### DIFF
--- a/src/UglyToad.PdfPig/Geometry/PdfPath.cs
+++ b/src/UglyToad.PdfPig/Geometry/PdfPath.cs
@@ -37,7 +37,7 @@ namespace UglyToad.PdfPig.Geometry
         }
 
         /// <summary>
-        /// Return true if points are organised with a clockwise order. Works only with closed paths.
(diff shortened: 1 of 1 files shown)


EliotJones:
Looks good to me, I'm not sure whether `IsClockwise`/`IsCounterClockwise` should be moved out to a utility method rather than living on `PdfPath` itself but I'm happy with it as-is for now. 

ToLines is a useful method for Bezier curves too, it's a good addition.

Thanks for the PR.

PR #69, opened 2019-10-07
Support for hORC, Atlo Xml and Page Xml output formats

BobLd:
The `AltoDocument` and `PageXmlDocument` classes are marked as `public` because it is mandatory in order to use `XmlSerializer.Serialize()`.
I've used the `[EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]` in order to hide them (the best would be to make the `internal`, but not possible). The objective being to avoid all these classes to create a mess :)

The `ITextExporter` implementations are marked as `internal` because they need more testing, but the export functions are available through `Page.Experimental`.

### Example
An example of all output formats can be found [here](https://github.com/UglyToad/PdfPig/files/3697901/text.samples.zip).

### Documentation
- see [here](https://en.wikipedia.org/wiki/HOCR) and [here](http://kba.cloud/hocr-spec/1.2/) for more about __hOCR__
- see [here](https://www.loc.gov/standards/alto/) and [here](https://en.wikipedia.org/wiki/ALTO_(XML)) for more about __Alto__
- see [here](https://github.com/PRImA-Research-Lab/PAGE-XML) for more about __Page Xml__
### Tested with:
- 'hocrjs' for hORC (see https://unpkg.com/hocrjs)
- 'PAGE Viewer' for hORC, Atlo Xml and Page Xml (see http://www.primaresearch.org/tools/PAGEViewer) (___NB: 'PAGE Viewer' apparently does not support the xml `Description` tag. Remove it from the xml file if you want to view it___)

### Preview
Preview of __hOCR__ output using [hocrjs](https://unpkg.com/hocrjs):
Capture2


src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+        {
+            return new AltoDocument()
+            {
+                Layout = new AltoDocument.AltoLayout()
+                {
+                    StyleRefs = null
+                },
+                Description = GetAltoDescription(fileName),
+                SchemaVersion = "4",
+                //Styles = new AltoStyles() { },
```
EliotJones: It's worth always removing commented code.

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+                //Tags = new AltoTags() { }
+            };
+        }
+
+        /// <summary>
```
EliotJones: I'd remove these empty doc comments, since they're not needed on private or internal members.

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+                    HPos = (float)Math.Round(rectangle.Value.Left * scale),
+                    Height = (float)Math.Round(rectangle.Value.Height * scale),
+                    Width = (float)Math.Round(rectangle.Value.Width * scale),
+                    Rotation = 0,
+                    //Cs = false,
+                    StyleRefs = null,
+                    TagRefs = null,
+                    title = null,
+                    type = null,
+                    //IdNext = "NA", // for reading order
```
EliotJones: More commented code to remove.

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+
+            return new AltoDocument.AltoIllustration()
+            {
+                VPos = (float)Math.Round((height - rectangle.Top) * scale),
+                HPos = (float)Math.Round(rectangle.Left * scale),
+                Height = (float)Math.Round(rectangle.Height * scale),
+                Width = (float)Math.Round(rectangle.Width * scale),
+                FileId = "",
+                Rotation = 0,
+                //IdNext = "NA", // for reading order
```
EliotJones: Commented code

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+                Id = "P" + pageCount + "_I" + illustrationCount.ToString("#00000")
+            };
+        }
+
+        /// <summary>
```
EliotJones: Empty private doc comments

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+                //Encoding = new System.Text.UTF8Encoding(true),
+                Indent = true,
+                IndentChars = indentChar,
+                OmitXmlDeclaration = true // hack to manually handle utf-8
+            };
+
+            using (var stringWriter = new System.IO.StringWriter())
+            using (var xmlWriter = XmlWriter.Create(stringWriter, settings))
+            {
+                stringWriter.WriteLine("<?xml version=\"1.0\" encoding=\"UTF-8\"?>"); // hack to manually handle utf-8
```
EliotJones: Could something like this work better for this purpose? https://timvw.be/2007/01/08/generating-utf-8-with-systemxmlxmlwriter/

Does the Alto format require UTF8?

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+    /// See https://github.com/altoxml/schema
+    /// </summary>
+    [EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+    [GeneratedCodeAttribute("xsd", "4.6.1055.0")]
+    [SerializableAttribute()]
+    [DebuggerStepThroughAttribute()]
+    [DesignerCategoryAttribute("code")]
+    [XmlTypeAttribute(Namespace = "http://www.loc.gov/standards/alto/ns-v4#")]
+    [XmlRootAttribute("alto", Namespace = "http://www.loc.gov/standards/alto/ns-v4#", IsNullable = false)]
+    public class AltoDocument
```
EliotJones: I'd move this to a new file and for each of the properties (except where required by the `isXXFieldSpecified` setters) change it to be an auto property, I'll add an example of this below...

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
```diff
@@ -0,0 +1,5337 @@
...
+        private AltoTags tagsField;
+
+        private AltoLayout layoutField;
+
+        private string sCHEMAVERSIONField;
+
+        /// <summary>
+        /// Describes general settings of the alto file like measurement units and metadata
+        /// </summary>
+        public AltoDescription Description
```
EliotJones: For example:

```suggestion
        public AltoDescription Description { get; set; }
```

And remove `private AltoDescription descriptionField;`.

Unless this is required by the XML serializer?

src/UglyToad.PdfPig/Content/Page.cs
```diff
@@ -171,6 +171,49 @@ public decimal GetPointSize(Letter letter)
             {
                 return letter.PointSize;
             }
+
```
EliotJones: I think rather than having these exporters live on the page API it's better for them to be standalone functionality that is then documented, either in the readme or the wiki. By standalone I mean the consumer can new up an instance where required, for example:

    // Constructor provides sensible defaults, or arguments are moved to the method call itself.
    ITextExporter hOcrExporter = new HOcrExporter();

It's a balance between discoverability vs making the public API too complex and I think for these export formats people are going to want to opt-in and will already know about them, whereas for most items on the public API of `Page` and `PdfDocument` they're designed so that people can autocomplete to victory. As long as we document the existence of these exporter types and provide easy on-ramps for using them, e.g.:

    hOcrExporter.Export(PdfDocument document, IPageSegmenter segmenter = null, IWordExtractor wordExtractor = null, ...);

Then people who are seeking HOcr, Alto and PageXml formats will be able to find and use them, whereas for people who are not familiar with those formats they won't be confused on encountering them on the page API.

src/UglyToad.PdfPig/Export/ITextExporter.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Exports the page's text into the desired format.
+    /// </summary>
+    public interface ITextExporter
+    {
+        /// <summary>
+        /// Get the text representation.
+        /// </summary>
+        /// <param name="page"></param>
```
EliotJones: I'd remove these empty doc comments since they're hopefully fairly self-explanatory.

src/UglyToad.PdfPig/Export/ITextExporter.cs
```diff
@@ -0,0 +1,17 @@
...
+        /// <returns></returns>
+        string Get(Page page);
+    }
```
EliotJones: I'd suggest a method on here for exporting the whole document, if that's a supported feature of these formats, or is each XML/HTML document only valid on a per page basis?

```suggestion
        string ExportDocument(PdfDocument document);
    }
```

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -211,6 +211,10 @@ internal void ClosePath()
             commands.Add(new Close());
         }
 
+        /// <summary>
+        /// The rectangle completely containing the path.
+        /// </summary>
+        /// <returns></returns>
```
EliotJones: Just another empty doc comment here.

src/UglyToad.PdfPig/Export/HOcrTextExporter.cs
```diff
@@ -0,0 +1,358 @@
...
+        public HOcrTextExporter(IWordExtractor wordExtractor, IPageSegmenter pageSegmenter, double scale = 1.0, string indent = "\t")
+        {
+            this.wordExtractor = wordExtractor;
+            this.pageSegmenter = pageSegmenter;
+            this.scale = (decimal)scale;
+            this.indentChar = indent;
+        }
+
+        /// <summary>
+        /// Get the hORC (HTML) string of the page layout.
```
EliotJones: I'd do a find and replace on `hORC <-> hOCR`, it's an easy mistake to make 😆

src/UglyToad.PdfPig/Export/PageXmlTextExporter.cs
```diff
@@ -0,0 +1,9820 @@
...
+    public class PageXmlDocument
+    {
+        private PageXmlMetadata metadataField;
+
+        private PageXmlPage pageField;
+
+        private string pcGtsIdField;
+
+        /// <remarks/>
+        public PageXmlMetadata Metadata
```
EliotJones: I'm just going to try forking this branch and using Resharper to auto-auto-property these or it's going to take forever 😆

src/UglyToad.PdfPig/Export/HOcrTextExporter.cs
```diff
@@ -0,0 +1,358 @@
...
+        /// </summary>
+        /// <param name="document">The document.</param>
+        /// <param name="includePaths">Draw <see cref="PdfPath"/>s present in the page.</param>
+        /// <param name="useHocrjs">Will add a reference to the 'hocrjs' script just before the closing 'body' tag, adding the 
+        /// interface to a plain hOCR file.<para>See https://github.com/kba/hocrjs for more information.</para></param>
+        public string Get(PdfDocument document, bool includePaths = false, bool useHocrjs = false)
+        {
+            string hocr = GetHead() + indentChar + "<body>\n";
```
EliotJones: I'm a bit worried about using string concatenation to generate XML, it seems like something that's easy to get wrong, it's probably OK as a first pass but I'm not sure how well it'll handle escape characters etc. I'd look into either using `XDocument` or `XmlSerializer` for this use case too, I know it's HTML so technically different but I think it's better to use something like that to generate well-formatted XML.

BobLd: Hi Eliot, agreed on this, but I have not found the hOCR schema, so I had to do it manually. This is the approach used in Tesseract as well for hOCR. See https://github.com/tesseract-ocr/tesseract/blob/master/src/api/hocrrenderer.cpp

If you are able to find the Schema, happy to look at it 😄

src/UglyToad.PdfPig/Export/HOcrTextExporter.cs
BobLd: hehe yes!

src/UglyToad.PdfPig/Export/ITextExporter.cs
BobLd: This is a good question and the answer is not clear to me. It seems PageXml only support per page basis, and hOCR and Alto support both. The problem with the whole document approach with all methods is that they use a reference image of the page, so that bounding boxes can be drawn on the picture (see the hOCR html file rendering with _hocrjs_). 

So the whole document approach will not be rendered correctly.

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
BobLd: It seems they all need UTF8. Will look at your link

src/UglyToad.PdfPig/Export/AltoXmlTextExporter.cs
BobLd: all the private fields were generated automatically...

src/UglyToad.PdfPig/Content/Page.cs
BobLd: okay makes sense, let's make them `public` and remove them from the `Page` 👍

BobLd:
I implemented UTF8, thanks for the link!

PR #72, opened 2019-10-15
Fix export formatting

EliotJones:
@BobLd just following on from #69 I've done some fairly intensive refactoring on AltoXml, I've moved them into a subfolder of exporters and split each partial class into its own file, fixing any naming issues where appropriate.

I've also renamed `DocstrumBB` to `DocstrumBoundingBoxes` to make the abbreviation clearer and done a lot of refactoring internally to appease Resharper. There was also a bug when using Docstrum on page 16 of the `PigProductionHandbook.pdf` which this refactor has fixed. It's worth testing Docstrum and AltoXml on this version just to make sure I've not broken anything!


BobLd:
Hi Eliot,

Thanks a lot for the huge work done, I'll do my best to check this before end of week!

BobLd:
I had a look and everything seems fine.

Thanks for your improvements on the `DocstrumBB`, it makes the code way more readable!

On the Alto side: seems very good! The only change I would make is to let the `AltoXmlTextExporter` outside of the Alto folder (or change the namespace from `UglyToad.PdfPig.Export.Alto` to `UglyToad.PdfPig.Export`) so that it is easier to find. 

Let me know what you think

EliotJones:
Thanks, I've moved the exporter back to the exporters namespace and also moved one of the methods from Docstrum to live on `PdfRectangle`. I'll merge this in now and probably do something similar for the PageXml exporter XML classes when I get time.

PR #92, opened 2019-12-05
Outlines refactoring

EliotJones:
@BobLd I've made a few changes here to preserve the immutability of public API members (`BookmarkNode`). I've split `BookmarkNode` into `DocumentBookmarkNode` and `ExternalBookmarkNode` to make the distinction explicit.

I've also wired up the `isLenientParsing` mode stuff to throw errors more often if the specification isn't matched.

Mainly I've changed everything to use the `TryGet` overloads of `DictionaryToken` which takes the `IPdfTokenScanner` and will handle locating tokens of the correct type for a `DictionaryToken`'s value. I also changed the recursive algorithm of reading the bookmark's tree because I worried it could end up with either an infinite loop or a very deep stack (since each sibling is a new stack frame then each child then each sibling of each child, etc). This made the code a lot more complicated so I need to test it a bit more.

I also load the named destinations tree once at the start of processing (and add support for the 1.1 specification) but I'm not sure if this was worth doing.


src/UglyToad.PdfPig/Outline/BookmarksProvider.cs
```diff
@@ -1,470 +1,442 @@
...
-                var pageNumber = structure.Catalog.GetPageByReference(pageIndirectReferenceToken.Data).PageNumber;
-                if (pageNumber.HasValue)
-                {
-                    currentNode.PageNumber = pageNumber.Value;
-                }
-                else
-                {
-                    log.Error("BookmarksProvider.GetDestination(): Cannot find page number.");
-                }
+                pageNumber = catalog.GetPageByReference(pageIndirectReferenceToken.Data).PageNumber ?? 1;
```
BobLd: Hi Eliot, 

thanks a lot for your help on that! Looks very good now.
Only comment I have at the moment is that some pdf files throw `NullReferenceException` (see below).

The easy fix is to replace 
`pageNumber = catalog.GetPageByReference(pageIndirectReferenceToken.Data).PageNumber ?? 1;` by 
`pageNumber = catalog.GetPageByReference(pageIndirectReferenceToken.Data)?.PageNumber ?? 1;`

Full error:
`
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=UglyToad.PdfPig
  StackTrace:
   at UglyToad.PdfPig.Outline.BookmarksProvider.TryGetExplicitDestination(ArrayToken explicitDestinationArray, Catalog catalog, ILog log, ExplicitDestination& destination) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 284
   at UglyToad.PdfPig.Outline.BookmarksProvider.TryReadExplicitDestination(IToken value, Catalog catalog, IPdfTokenScanner pdfScanner, ILog log, ExplicitDestination& destination) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 257
   at UglyToad.PdfPig.Outline.BookmarksProvider.ExtractNameTree(DictionaryToken nameTreeNodeDictionary, Catalog catalog, IPdfTokenScanner pdfScanner, Boolean isLenientParsing, ILog log, Dictionary2 explicitDestinations) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 223
   at UglyToad.PdfPig.Outline.BookmarksProvider.ExtractNameTree(DictionaryToken nameTreeNodeDictionary, Catalog catalog, IPdfTokenScanner pdfScanner, Boolean isLenientParsing, ILog log, Dictionary2 explicitDestinations) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 236
   at UglyToad.PdfPig.Outline.BookmarksProvider.ReadNamedDestinations(Catalog catalog, IPdfTokenScanner pdfScanner, Boolean isLenientParsing, ILog log) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 201
   at UglyToad.PdfPig.Outline.BookmarksProvider.GetBookmarks(Catalog catalog) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\Outline\BookmarksProvider.cs:line 47
   at UglyToad.PdfPig.PdfDocument.TryGetBookmarks(Bookmarks& bookmarks) in D:\VS2017\source\repos\BobLd\PdfPig\src\UglyToad.PdfPig\PdfDocument.cs:line 215
`

EliotJones: Thanks, I've changed this to skip explicit destinations where the page specifies a reference that doesn't exist in the page tree instead of defaulting to page 1, not sure if this will cause missing bookmarks though, would you be able to test the newest version just to check please?

BobLd: I've tested it and it seems to work properly, but agree with on the missing bookmarks question. 

As an aside, if we ever need to default to a page number, I would chose -1 to signal that this is not an actual page.

EliotJones: Thanks for testing that.

That makes sense, the reason I've avoided using the `-1` or `0` approach at the moment is it relies on the consumer knowing that they need to check the page number for some edge condition where it won't be a valid index, for that reason I'd prefer to use `int?` which at least makes the edge condition explicit (though to be fair the -1 approach is fairly common in `IndexOf` etc). But for now I'd rather skip the bookmark entirely, I think there may be a case where the page indirect reference links to the real indirect reference for the page object but I'll merge this for now and squish any bugs as and when they're encountered.

PR #100, opened 2019-12-10
Fix issue with leaf/nodes in RecursiveXYCut that only have whitespaces words.

huzhiguan:
In VerticalCut and HorizontalCut, it's using `CountWords() <= 1` as part of the condition to stop cutting with small text segments. This function doesn't exclude all the whitespace words. I made changes to make sure that those two functions will return when finding  XYLeaf/XYNode only have whitespaces words.

In `GetBlocks()`, if the returned node is an XYLeaf node, the `node.GetLeafs()` will return null which will cause the following lines to throw an exception.  I modified that part to make sure it's handling leaf node differently.

### DIFF
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
@@ -70,11 +70,18 @@ namespace UglyToad.PdfPig.DocumentLayoutAnalysis
             XYLeaf root = new XYLeaf(pageWords); // Create a root node.
             XYNode node = VerticalCut(root, minimumWidth, dominantFontWidthFunc, dominantFontHeightFunc);
 
-            var leafs = node.GetLeafs();
-
-            if (leafs.Count > 0)
+            if (node.IsLeaf)
             {
-                return leafs.Select(l => new TextBlock(l.GetLines())).ToList();
+                return new List<TextBlock>{ new TextBlock((node as XYLeaf).GetLines())};
+            }
+            else
+            {
+                var leafs = node.GetLeafs();
+
+                if (leafs.Count > 0)
+                {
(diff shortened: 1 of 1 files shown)


src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -84,6 +91,18 @@ public IReadOnlyList<TextBlock> GetBlocks(IEnumerable<Word> pageWords, decimal m
...
+
+            if(!words.Any())
+            {
+                return new XYNode(null);
+            }
+            else
+            {
+                //Create new leaf with non-whitespace words.
+                leaf = new XYLeaf(words);
+            }
```
BobLd: Go to line here

src/UglyToad.PdfPig/DocumentLayoutAnalysis/RecursiveXYCut.cs
```diff
@@ -177,14 +193,25 @@ public IReadOnlyList<TextBlock> GetBlocks(IEnumerable<Word> pageWords, decimal m
...
+
             if (leaf.CountWords() <= 1)
             {
                 // we stop cutting if 
                 // - only one word remains
                 return leaf;
             }
 
-            var words = leaf.Words.Where(w => !string.IsNullOrWhiteSpace(w.Text)).OrderBy(w => w.BoundingBox.Bottom).ToArray(); // order bottom to top
```
BobLd: Remove line here

BobLd (changes requested):
Some formatting comments

BobLd (approved):
Thanks for your help on that!

huzhiguan:
> Thanks for your help on that!

You're welcome. Thank you all for the great project!

EliotJones:
Thank you for the contribution and sorry for the delay in merging.

PR #102, opened 2019-12-14
Improving Geometry classes with Tests

BobLd:
- Implementing `Equals()` and `GetHashCode()` for all paths (here let me know what you think, especially on the Hash)

- PdfPoint:
	- MoveXY
	
- PdfRectangle:
	- Contains
	- IntersectsWith
	- Intersect
	- MoveXY
	
- PdfLine:
	- GetBoundingRectangle
	- Contains
	- IntersectsWith
	- Intersect
	- ParallelTo
	
- Line:
	- Contains
	- IntersectsWith
	- Intersect
	- ParallelTo
	
- Bezier Curve:
	- Split (see _Computer Aided Geometric Design_ by T. W. Sederberg, section 2.3)
	- Intersect
	- FindIntersectionT

I've also put some test, but we would need some for the bezier curve

### DIFF
--- a/src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
+++ b/src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
@@ -1,0 +1,508 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using static UglyToad.PdfPig.Geometry.PdfPath;
+
+namespace UglyToad.PdfPig.Geometry
+{
+    /// <summary>
+    /// Extension class to Geometry.
+    /// </summary>
+    public static class GeometryExtensions
+    {
+        #region PdfRectangle
+        /// <summary>
+        /// Whether the rectangle contains the point.
+        /// <para>Return false if the point belongs to the border.</para>
+        /// </summary>
+        public static bool Contains(this PdfRectangle rectangle, PdfPoint point)
+        {
+            return point.X > rectangle.Left &&
+                   point.X < rectangle.Right &&
+                   point.Y > rectangle.Bottom &&
+                   point.Y < rectangle.Top;
+        }
+
+        /// <summary>
+        /// Whether two rectangles overlap.
+        /// <para>Returns false if the two rectangles only share a border.</para>
+        /// </summary>
+        public static bool IntersectsWith(this PdfRectangle rectangle, PdfRectangle other)
+        {
+            if (rectangle.Left > other.Right || other.Left > rectangle.Right)
+            {
+                return false;
+            }
+
+            if (rectangle.Top < other.Bottom || other.Top < rectangle.Bottom)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Get the <see cref="PdfRectangle"/> that is the intersection of two rectangles.
+        /// </summary>
+        public static PdfRectangle? Intersect(this PdfRectangle rectangle, PdfRectangle other)
+        {
+            if (!rectangle.IntersectsWith(other)) return null;
+            return new PdfRectangle(Math.Max(rectangle.BottomLeft.X, other.BottomLeft.X),
+                                    Math.Max(rectangle.BottomLeft.Y, other.BottomLeft.Y),
+                                    Math.Min(rectangle.TopRight.X, other.TopRight.X),
+                                    Math.Min(rectangle.TopRight.Y, other.TopRight.Y));
+        }
+        #endregion
+
+        #region PdfLine
+        /// <summary>
+        /// Whether the line segment contains the point.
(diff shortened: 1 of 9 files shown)


src/UglyToad.PdfPig/Geometry/PdfPoint.cs
```diff
@@ -77,6 +77,17 @@ public PdfPoint MoveY(decimal dy)
             return new PdfPoint(X, Y + dy);
         }
 
+        /// <summary>
+        /// Creates a new <see cref="PdfPoint"/> which is the current point moved in the x and y directions relative to its current position by a value.
+        /// </summary>
+        /// <param name="dx">The distance to move the point in the x direction relative to its current location.</param>
+        /// <param name="dy">The distance to move the point in the y direction relative to its current location.</param>
+        /// <returns>A new point shifted on the y axis by the given delta value.</returns>
+        public PdfPoint MoveXY(decimal dx, decimal dy)
```
EliotJones: It might be worth naming this `Translate` which is the name used for the same method for `TransformationMatrix`.

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -18,6 +18,11 @@ public class PdfPath
         /// </summary>
         public IReadOnlyList<IPathCommand> Commands => commands;
 
+        /// <summary>
+        /// True if the <see cref="PdfPath"/> was originaly draw as a rectangle.
+        /// </summary>
+        internal bool IsDrawnAsRectangle { get; set; }
```
EliotJones: I'm fine with this appearing on the public API since it's useful information. A better name might be `IsPartOfRectangle` maybe? Not really sure. (also remove the public setter please :) )

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -638,6 +714,38 @@ internal void Rectangle(decimal x, decimal y, decimal width, decimal height)
             LineTo(x + width, y + height);
             LineTo(x, y + height);
             LineTo(x, y);
+            IsDrawnAsRectangle = true;
+        }
+
+        /// <summary>
+        /// Order matters
```
EliotJones: I'd add a bit of extra explanation here, e.g. `Compares 2 PdfPaths for equality. Paths will only be considered equal if the commands which construct the paths are in the same order`. And also an `<inheritdoc/>`.

src/UglyToad.PdfPig/Geometry/PdfPath.cs
```diff
@@ -638,6 +714,38 @@ internal void Rectangle(decimal x, decimal y, decimal width, decimal height)
...
+        /// </summary>
+        public override int GetHashCode()
+        {
+            var hash = this.Commands.Count + 1;
+            for (int i = 0; i < this.Commands.Count; i++)
+            {
+                hash = hash * (i + 1) + this.Commands[i].GetHashCode();
```
EliotJones: All the `GetHashCode` implementations in this file look good but this one makes me unsure. I think it's probably 🤞 ok but it might be worth using something from here? https://stackoverflow.com/questions/3404715/c-sharp-hashcode-for-array-of-ints

src/UglyToad.PdfPig/Geometry/PdfPath.cs
BobLd: Agree on making it `public` but I would keep the name as only a rectangle (and not a part) can have this flag set a `true`. (Especially true for clipping rectangles)

src/UglyToad.PdfPig/Geometry/PdfPath.cs
BobLd: Agreed. just not sure what you mean regarding the `<inheritdoc/>` :D

src/UglyToad.PdfPig/Geometry/PdfPath.cs
BobLd: I've modified it slightly but I basically kept the same logic (cf accepted answer in your link). Let me know if you think of something else :)

src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
```diff
@@ -0,0 +1,508 @@
...
+{
+    /// <summary>
+    /// Extension class to Geometry.
+    /// </summary>
+    public static class GeometryExtensions
+    {
+        #region PdfRectangle
+        /// <summary>
+        /// Whether the rectangle contains the point.
+        /// <para>Return false if the point belongs to the border.</para>
```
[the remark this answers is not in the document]
BobLd: Agreed, I've added a flag

Issue #104, opened 2019-12-16
Parallel.For in DocumentLayoutAnalysis


huzhiguan:
I noticed that inside namespace `DocumentLayoutAnalysis`, some algorithms are using `Parallel.For` to accelerate processing speed. From a user's standpoint, would it be better to get more visibility and control over the thread usage? For example, a place we can set `MaxDegreeOfParallelism` when needed. Thanks!

BobLd:
Hi, that is a good point and will be useful. I should find some time to work on it soon.

BobLd:
Hi, can you have a look at #105? Let me know if you have any comment

huzhiguan:
> Hi, can you have a look at #105? Let me know if you have any comment

It looks good to me, thanks for the quick update!

EliotJones:
Hi, this should be available via Nuget in the latest 0.1.0 release, are you happy to close this now?

PR #105, opened 2019-12-18
Whitespace covering algorithm and #104

BobLd:
Adding Whitespace covering algorithm and adding support for `MaxDegreeOfParallelism` in DocumentLayoutAnalysis (cf. #104)

#### References
- Section 3.2 of [_High precision text extraction from PDF documents_](https://www.duo.uio.no/bitstream/handle/10852/8893/Berg.pdf?sequence=1) by Øyvind Raddum Berg 
- Section 2 of [_Two geometric algorithms for layout analysis_](https://www.researchgate.net/publication/2504221_Two_Geometric_Algorithms_for_Layout_Analysis) by Thomas M. Breuel

Comments in the implementation are taken from the first reference.

#### Result example:
whitespace covering


EliotJones:
This looks great and I'm happy to merge it in. One thing I was wondering though from the Berg paper in 3.2.6 he mentions strategies to avoid intraparagraph lines. It seems like the current approach produces a fair few of them, for example:

image

Did experimenting with a minimum height (perhaps based on font-height/size metrics) for wide rectangles result in missing genuine covering rectangles or is it a future enhancement of the algorithm?

BobLd:
Hi Eliot,

I think the `minWidth` and `minHeight` parameters are what you are looking for. The default values are:
```
public static IReadOnlyList<PdfRectangle> GetWhitespaces(IEnumerable<Word> words, IEnumerable<IPdfImage> images, int maxRectangleCount = 40, int maxBoundQueueSize = 0)
{
	return GetWhitespaces(words,
	  images,
	  words.SelectMany(w => w.Letters).Select(x => x.GlyphRectangle.Width).Mode() * 1.25m, // <- here
	  words.SelectMany(w => w.Letters).Select(x => x.GlyphRectangle.Height).Mode() * 1.25m, // <- here
	  maxRectangleCount: maxRectangleCount,
	  maxBoundQueueSize: maxBoundQueueSize);
}
```

From the paper:
> This was solved by introducing lower bounds for the height and width of rectangles.
As an optimization, conservative page-wide bounds were precalculated based on average
font sizes (for vertical bounding) and average vertical distance between characters (for
horizontal bounding).

Berg uses average and I use mode, happy to change that if you want :) but this can be tailored by the user anyway. Or maybe we can use 1.5 or 2 times the mode Width and Height instead of 1.25.

I didn't try to optimise too much the intraparagraph lines problem because the following step of the process (not implemented here) is about looking for column separators, and these lines would be filtered out.

Let me know if I missed something

BobLd:
Below are some tests:
- Using `1.5 * Mode()`
Random 2 Columns Lists Images 1constrained_textline_1 50_small

- Using `2.0 * Mode()`
Random 2 Columns Lists Images 1constrained_textline_2 00_small

Issue #47, opened 2019-07-23
Relatively slow processing


vadik299:
This is not exactly an issue, but more like a general question. While extracting Letters collection I noticed that overall the process runs about 4-5 times slower than PdfBox. I run PdfBox through Ikvm, so I was expecting it to be the other way around :-)
Of course there can be many things contributing to this, but I did one quick test - I ran a mass replace of word "decimal" to "double" across the whole code base. And yes, the speed got right on par with PdfBox! Changing it to "float" made it even a little faster (probably due to smaller memory footprint).
Sure double/float is not precise, but personally I often need to run extraction over hundreds of thousands  PDFs, so speed is crucial and the time difference is substantial. On the other hand I think that letters/lines positions and dimensions would be OK with 2 digits precision at most (ok, maybe 3:-))
I see a few ways to approach this:
1. Change all properties to float. This is quick, but not very user friendly and may have some minor issues if anyone ever tries to compare numbers directly.
2. Keep values internally as Int multiplied by 100 (or 1000). Do all calculations on Int, then return to the user as decimal (divide by 100) - this may be better, but probably harder to do and potentially more confusing.
3. Keep everything as is and suggest getting a better server :-) (or optimize somewhere else)

Thoughts? Thank you for your time and all the work put in this library!

EliotJones:
Thanks for the feedback, it's useful to know the performance characteristics in production. 

I'm surprised it's so much slower! I'm going to run the profiling this weekend but it sounds like the decimals are a large part of it.

The reason for using decimals over doubles initially is that it allows accurate representation of the numbers as they appear in the document, so for example a font size is reported as 12.5 rather than 12.499999 or whatever else doubles get up to.

It seems like where multiple operations are being done on those decimal values though (matrix transformations for processing a page content stream) they could get away with being doubles or even floats then cast back to decimals on the public API. This keeps decimals on the public facing API keeping the values true to what was reported on the face of the document for things like writing or reading numeric values, but speeds up processing which loses some accuracy anyway.

[...]

PR #106, opened 2019-12-21
use doubles where calculations are being run

EliotJones:
for #47 it seems that decimals are still too slow after all other performance optimizations. this change aims to preserve decimals where the value being read is exactly representable (all numeric tokens in the document, font metrics files, some fonts and page content stream) but use doubles where the value is calculated, e.g. transformation matrices, etc.

One place where I'd like to maintain decimals is for `Page.Width` and `Page.Height` since these are exact, not calculated values.

The other outstanding performance optimization is to add parsing settings to control the amount of data retrieved on opening a page. Currently we populate the `Operations` array with all operations parsed from the content stream, but most users don't need access to these where they're just interested in text extraction (I assume).

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Geometry/PdfLineTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Geometry/PdfLineTests.cs
@@ -21,50 +21,50 @@ namespace UglyToad.PdfPig.Tests.Geometry
         public void Length()
         {
             var line = new PdfLine(2, 1, 6, 4);
-            Assert.Equal(5m, line.Length);
+            Assert.Equal(5d, line.Length);
 
             var line2 = new PdfLine(-2, 8, -7, -5);
-            Assert.Equal(13.93m, Math.Round(line2.Length, 2));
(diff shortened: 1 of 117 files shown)


src/UglyToad.PdfPig.Tests/Geometry/PdfLineTests.cs
```diff
@@ -93,81 +93,5 @@ public void ParallelTo()
             Assert.True(obliqueLine1.ParallelTo(obliqueLine2));
             Assert.True(obliqueLine2.ParallelTo(obliqueLine1));
         }
-
-        [Fact]
-        public void IntersectsWithLine()
-        {
-            var verticalLine1 = new PdfLine(10, 7.5m, 10, 15);
```
EliotJones: @BobLd these tests seemed to start failing after changing `PdfLine` and `PdfPoint` to use doubles instead of decimals. I've deleted them for now because I assume it was an accuracy issue.

BobLd: Hi Eliot, no problem I'll fixe them when I have time

PR #115, opened 2019-12-31
Add headers/footers (decoration) classifier for TextBlock

BobLd:
## `MinimumEditDistance`
_The minimum edit distance between two strings is the minimum number of editing operations: Insertion, Deletion, Substitution, needed to transform one into the other._

### Reference
[_Minimum Edit Distance_ presentation](https://web.stanford.edu/class/cs124/lec/med.pdf) by Dan Jurafsky

## `DecorationTextBlockClassifier`
Algorithm that retrieve blocks that are labelled as decoration (e.g. headers, footers) for each page in the document, using a content and a geometric similarity measure.
Decoration blocks are blocks that contains information such as author names, publication titles, page numbers, etc. They are printed repeatedly at the border of each page, usually placed inside headers or footers, but sometimes also at the left or right edge of the page.</para>

Comments and Summaries in the code are taken from the paper in reference.

### Reference
[_Unsupervised document structure analysis of digital scientific articles_](http://www.know-center.tugraz.at/download_extern/papers/ijdl-2013.pdf) by S. Klampfl, M. Granitzer, K. Jack, R. Kern.

### Usage
```csharp
  var decorations = DecorationTextBlockClassifier.Get(document.GetPages().ToList(),
                        DefaultWordExtractor.Instance,
                        RecursiveXYCut.Instance);
```

### Example
Highlighted in blue are text blocks classified as decoration (from the [FarmerMac.pdf](https://github.com/UglyToad/PdfPig/blob/master/src/UglyToad.PdfPig.Tests/Integration/Documents/FarmerMac.pdf)):
FarmerMac 1

### DIFF
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/DecorationTextBlockClassifier.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/DecorationTextBlockClassifier.cs
@@ -19,10 +19,7 @@ namespace UglyToad.PdfPig.DocumentLayoutAnalysis
     /// </summary>
     public static class DecorationTextBlockClassifier
     {
-        /// <summary>
-        /// 
-        /// </summary>
-        public static Regex numbersPattern = new Regex(@"(\d+)|(\b([MDCLXVI]+)\b)", RegexOptions.IgnoreCase);
+        private static Regex numbersPattern = new Regex(@"(\d+)|(\b([MDCLXVI]+)\b)", RegexOptions.IgnoreCase);
         private static string replacementChar = "@";
 
         /// <summary>
--- a/src/UglyToad.PdfPig/DocumentLayoutAnalysis/DecorationTextBlockClassifier.cs
+++ b/src/UglyToad.PdfPig/DocumentLayoutAnalysis/DecorationTextBlockClassifier.cs
@@ -1,0 +1,267 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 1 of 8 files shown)


BobLd:
Hi Eliot, I just noticed that previous commit are still present in this PR... Let me know if this is an issue

src/UglyToad.PdfPig/DocumentLayoutAnalysis/DecorationTextBlockClassifier.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// <param name="n">Number of blocks in a page to be considered when looking for decoration blocks.</param>
+        /// <param name="maxDegreeOfParallelism">Sets the maximum number of concurrent tasks enabled. 
+        /// <para>A positive property value limits the number of concurrent operations to the set value. 
+        /// If it is -1, there is no limit on the number of concurrently running operations.</para></param>
+        public static IReadOnlyList<IReadOnlyList<TextBlock>> Get(IReadOnlyList<IReadOnlyList<TextBlock>> pagesTextBlocks,
+            Func<string, string, double> minimumEditDistanceNormalised, double similarityThreshold = 0.25, int n = 5, int maxDegreeOfParallelism = -1)
+        {
+            if (pagesTextBlocks.Count < 2)
+            {
+                throw new ArgumentException("The algorithm cannot be used with a document of less than 2 pages.", nameof(pagesTextBlocks));
```
EliotJones: Might be least surprising here to return an empty list instead of throwing. While both make sense I can see a scenario where someone is automating the processing of many documents without knowledge of their length or content and they can more easily handle an empty list, which is a valid return for documents of any length, than an exception.

PR #117, opened 2020-01-04
refactor font project

EliotJones:
creates 2 new projects to make the apis used more granular and support a separate font project for truetype, afm and adobe type 1 fonts.


EliotJones:
The resulting project structure is as follows:

image

PR #120, opened 2020-01-07
Fix for rectangle width/height incorrectly parsed

vadik299:


EliotJones:
Thanks, this is a useful fix.

PR #121, opened 2020-01-08
Add basic marked content extraction capabilities

BobLd:
Extract marked content and artifact marked content. Mostly follows PdfBox approach.

### Usage
```csharp
var mc = page.GetMarkedContents();
```


BobLd:
Close with #122

PR #130, opened 2020-01-31
Improve bounding box for word

BobLd:
- Improve bounding box for `Word`. The new method works as follow:
       - fit a line through the base line's points to get the slope
       - rotate the points to obtain the axis-aligned bounding box (AABB)
       - rotate back the AABB to obtain the oriented bounding box (OBB)

__Results:__
In dark (light) red, the old (new) algo:

image

- Implement the method above in `GeometryExtensions`
- Implement `TransformationMatrix.Inverse()`, needed in the above for the _rotate back_ step. Reference [here](https://en.wikipedia.org/wiki/Invertible_matrix#Inversion_of_3_%C3%97_3_matrices)

The new bounding box approach is inspired by https://arxiv.org/abs/1907.03892, in way much simpler:

image

### DIFF
--- a/src/UglyToad.PdfPig.Core/TransformationMatrix.cs
+++ b/src/UglyToad.PdfPig.Core/TransformationMatrix.cs
@@ -360,6 +360,30 @@
                 E * scalar, F * scalar, row3 * scalar);
         }
 
+                /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public TransformationMatrix Inverse()
+        {
+            var a = (D * row3 - row2 * F);
+            var c = -(C * row3 - row2 * E);
+            var e = (C * F - D * E);
+
+            var b = -(B * row3 - row1 * F);
+            var d = (A * row3 - row1 * E);
+            var f = -(A * F - B * E);
+
+            var r1 = (B * row2 - row1 * D);
+            var r2 = -(A * row2 - row1 * C);
+            var r3 = (A * D - B * C);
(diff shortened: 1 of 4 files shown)


src/UglyToad.PdfPig.Core/TransformationMatrix.cs
```diff
@@ -360,6 +360,30 @@ public TransformationMatrix Multiply(double scalar)
                 E * scalar, F * scalar, row3 * scalar);
         }
 
+                /// <summary>
```
EliotJones: It's worth putting a comment in here (and deleting the `returns` line).

src/UglyToad.PdfPig.Core/TransformationMatrix.cs
```diff
@@ -360,6 +360,30 @@ public TransformationMatrix Multiply(double scalar)
...
+            var f = -(A * F - B * E);
+
+            var r1 = (B * row2 - row1 * D);
+            var r2 = -(A * row2 - row1 * C);
+            var r3 = (A * D - B * C);
+            var det = A * a + B * c + row1 * e;
+
+            return new TransformationMatrix(a, b, r1, 
+                c, d, r2, 
+                e, f, r3).Multiply(1.0 / det);
```
EliotJones: I'd do the multiplication to each of the terms when calling the constructor instead of using the `Multiply` method call, this will avoid the need to create an unused matrix.

PR #134, opened 2020-02-09
Fix #133 Marked content extraction issue

### LINKED ISSUE #133: Marked content extraction issue
The `var markedContents = page.GetMarkedContents();` does not respect the tree structure: 
- All different marked content levels are treated as root level
- The index is not incremented properly

I'll fix the code soon

BobLd:
Fix #133 so that `var markedContents = page.GetMarkedContents();` respect the tree structure.


EliotJones:
If possible would you be able to add a test case for this please? Probably worth having some tests around it to check I don't break it in future. You can add a PDF document to test against to the `src\UglyToad.PdfPig.Tests\Integration\SpecificTestDocuments` directory then load it in and check the marked contents are populated in the same sort of way as loading files is done in, for example, `EncryptedDocumentTests.cs`.

BobLd:
Also fixed a bounding box extraction issue for artifact, like in the `Multiple Page - from Mortality Statistics.pdf` document.

PR #136, opened 2020-02-16
Faster clustering with kd-tree

BobLd:
Implement **kd-tree** data structure to make clustering faster. This will impact both `NearestNeighbourWordExtractor` and `DocstrumBoundingBoxes` classes as they use `ClusteringAlgorithms.ClusterNearestNeighbours(PdfPoint)`.

The kd-tree structure is only used in `ClusteringAlgorithms.ClusterNearestNeighbours(PdfPoint)` for the moment and not in `ClusteringAlgorithms.ClusterNearestNeighbours(PdfLine)`.

See the [wikipedia page](https://en.wikipedia.org/wiki/K-d_tree) for more informations on kd-trees:
- The cost of building the tree in is `O(n log n)` time in the best case. An algorithm that builds a balanced k-d tree to sort points has a worst-case complexity of `O(kn log n)`,
- Finding the nearest point is an `O(log n)` operation on average.

I ran performance tests to compare the current (`master`) implementation of `NearestNeighbourWordExtractor` and the one using the kd-tree (`kd-tree`): **203** documents were analysed for a total of **12,669** pages.

**The results show that the kd-tree implementation is 3.50 times faster on average.**
|    | master | kd-tree |
|---|:---:|:---:|
| **total time (min)** |16.5|4.7|
| **avg time/page (ms)** |78|22|

Below is a chart that plots the time spent on each page, as a function of the number of input letters:
kd tree improvement

### DIFF
--- a/src/UglyToad.PdfPig.DocumentLayoutAnalysis/KdTree.cs
+++ b/src/UglyToad.PdfPig.DocumentLayoutAnalysis/KdTree.cs
@@ -1,0 +1,230 @@
+﻿namespace UglyToad.PdfPig.DocumentLayoutAnalysis
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using UglyToad.PdfPig.Core;
+
+    // for kd-tree with line segments, see https://stackoverflow.com/questions/14376679/how-to-represent-line-segments-in-kd-tree 
+
+    internal class KdTree : KdTree<PdfPoint>
+    {
+        public KdTree(PdfPoint[] candidates) : base(candidates, p => p)
+        { }
+
+        public PdfPoint FindNearestNeighbours(PdfPoint pivot, Func<PdfPoint, PdfPoint, double> distanceMeasure, out int index, out double distance)
+        {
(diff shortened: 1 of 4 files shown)


src/UglyToad.PdfPig.DocumentLayoutAnalysis/KdTree.cs
```diff
@@ -0,0 +1,230 @@
...
+            /// <summary>
+            /// Split value.
+            /// </summary>
+            public double L => Depth == 0 ? Value.X : Value.Y;
+
+            public PdfPoint Value { get; }
+
+            public KdTreeNode<Q> LeftChild { get; internal set; }
+
+            public KdTreeNode<Q> RightChild { get; internal set; }
```
EliotJones: Can these setters be private / get only?

BobLd: Hi Eliot, we will need internal setters if we want to add an 'Add()' function to the tree. Good for me to leave it like this. Let me know

EliotJones: Sure that's fine, I'll merge it in.

PR #138, opened 2020-02-19
GeometryExtensions tests

BobLd:
Add several tests for the `GeometryExtensions` class, fix `PdfRectangle.GetWidthHeight()` and fix `BezierCurve.Intersect()`.


BobLd:
Some notes on `PdfRectangle` extensions. The following modifications were needed in order to account for rotated rectangles:
- `Contains()`
- `IntersectsWith()` (could be optimised, but works)

The `Intersect()` function only works for **not** rotated rectangles. The problem with rotated rectangles is that the intersection might not be a rectangle, see below for example. This will need to be tackled later on.
interesection

src/UglyToad.PdfPig.Tests/Geometry/PdfLineTests.cs
```diff
@@ -21,11 +24,315 @@ public void OriginIsZero()
         [Fact]
         public void Length()
         {
-            var line = new PdfLine(2, 1, 6, 4);
-            Assert.Equal(5d, line.Length);
+            Assert.Equal(578.0238169967487, new PdfLine(772.7935893651529, 299.87394515637465, 233.21554238153897, 92.58896203207878).Length, 6);
```
EliotJones: For most of these Geometry tests it'd be better to move all this repeated test data to a `Theory` using `MemberData`: http://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/

BobLd: Okay good point 😄 I did it it this way because the tests were generated via code. Will look at it!

src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
```diff
@@ -11,6 +11,19 @@
...
     {
+        private const double epsilon = 1e-5;
+
+        /// <summary>
+        /// Return true if the points are in counter-clockwise order.
+        /// </summary>
+        /// <param name="point1">The first point.</param>
+        /// <param name="point2">The second point.</param>
+        /// <param name="point3">The third point.</param>
+        private static bool ccw(PdfPoint point1, PdfPoint point2, PdfPoint point3)
```
EliotJones: A better name for this would be `IsCounterClockwise`.

src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
```diff
@@ -54,7 +67,7 @@ public static PdfPoint Subtract(this PdfPoint point1, PdfPoint point2)
         /// </param>
         internal static PdfRectangle ParametricPerpendicularProjection(IReadOnlyList<PdfPoint> polygon)
         {
-            if (polygon == null || polygon.Count == 0)
+            if (polygon == null || !polygon.Any())
```
EliotJones: Just as a general performance pointer, I'd avoid LINQ wherever you can, while it's more readable in this case it's very easy to end up accidentally putting LINQ on a hot path and clobbering perf. For instance I wrote a quick timing test for `Any` versus `Count` including lists with 0 and 1 items. The stats for 700 runs each looped 100,000 times were:

```
Count: Min: 3618        Max: 36413      Avg: 8659.79
  Any: Min: 15632       Max: 64904      Avg: 23914.74
```
Where the values are in ticks. It's possible I made an error writing this test and should have just used benchmark dot net but I think in the general case you're going to see worse performance with LINQ vs the 'naive' code alternative wherever it is used.

Now it doesn't matter too much here since this isn't a hot path as far as I'm aware but avoiding linq is a habit I like to keep to.

BobLd:
Hi eliot, do you know the best way to use `MemberData` for `IntersectsWith()` for example? How do you deal with  `Assert.True/False`?

EliotJones:
I'd use something like:

```
public IEnumerable<object[]> MyTestData
{
   new object[][] {
        new object[] { new PdfRectangle(etc), new PdfRectangle(etc2), true }
    }
}

[Theory]
[MemberData(nameof(MyTestData))]
public void IntersectsWith(PdfRectangle source, PdfRectangle other, bool expected)
{
    var result = source.IntersectsWith(other);
    Assert.Equal(expected, result);
}
```

I know in general one should use the specific bool asserts for XUnit but in this case it's fine to use equal for dynamic data.
```

src/UglyToad.PdfPig/Geometry/GeometryExtensions.cs
BobLd: I would leave it like this as it is the usual way this function is called in computational geometry.

PR #140, opened 2020-02-23
Add clipping intersect operation flag

BobLd:
- Add `ClippingRule` and `IsClipping` to `PdfPath`.
- Add `PdfPath.IsCounterClockwise` tests.


BobLd:
Hi Eliot, I just notice that ` internal void SetWindingRuleMode(int windingRule) { }` was present in `PdfPath` but I havn't used it, because it is `internal`. Let me know what you think

EliotJones:
Looks good to me, I'll merge it in and remove the old `SetWindingRuleMode` since it seems to be unused.

PR #144, opened 2020-02-29
Basic PdfMerger

InusualZ:
This class is a basic implementation of merging two documents together. The method that I am using is quite easy to understand. 

I'm creating a new `/Pages` Object for each new document that is being added. That way we avoid having to rename resources that may conflict with other document. We go through every `/Page` object in the root `/Pages` object of the document, copying and updating references of each object that is needed. This method is great, but has his pros and cons. 

#### **Pros**
- Fast - We copy only what the pages are explicitly requesting as needed
#### **Cons**
- Bigger Binaries - Due to we copying resources blindly (without checking if they are duplicated)

**Few thing that we can do**:
- We can analyze what we copy and de-duplicate it
- We can compress some resources (Ex. images)
- We can group resource they are referenced once

This optimization are great, but I thing they are better suited for later PR. They need some though before we implement them. We can even implement some of this feature to the creation of document (compress resources) or as standalone feature (sort of Pdf Optimizer)

PD: I want to thank @EliotJones for this project and @BobLd for the amazing work that's doing in the content extraction part of the project. I hope that one day my contributions are as great as yours.


EliotJones:
All looks good to me, I've pushed a few commits just to match project conventions and add the copy full node back to `/Pages` but otherwise it's great.

I thought of a couple of additional things while I was checking it that might need to be addressed in future.

1. Object streams. In general objects in the PDF document are of the `1 0 obj <content> endobj` form but they can also appear in streams, starting in PDF 1.7 I think. I've no idea what will happen at this stage if we try merging those. The document `Single Page Simple - from google drive.pdf` in the integration documents is an example of a document containing an object stream. I'm happy with not supporting them at the moment though and adding it in later.
2. Objects sharing an indirect reference. I'm not sure what would happen if 2 objects shared the same indirect reference, I think currently the referenced object would be duplicated, which is fine. For example in the following situation:

```
1 0 obj
[ 0 0 700 600 ]
endobj

2 0 obj
<</Type /Page /MediaBox 1 0 R ... >>
endobj

3 0 obj
<</Type /Page /MediaBox 1 0 R ...>>
endobj
```

I think the resulting document contains 2 copies of the object `1 0` but I'm not certain that is what happens. This was probably related to your first bulletpoint in things we can do for the file size. For now it's fine but I will need to try and create a test document to confirm this behaviour in future.

PR #150, opened 2020-03-08
Merger API: Lower object count

InusualZ:
First attempt at trying to lower the binary size of merged document.  The first thing that I did was to 
create a heavy modified version of `UglyToad.PdfPig.Writer.Fonts.BuilderContext` as `UglyToad.PdfPig.Writer.PdfStreamWriter`. This class would be in charge of writing tokens to an arbitrary stream. Token would only be written once, as we push token to the stream, the class would make sure to look for token already written, returning an `IndirectReferenceToken` to it.

When we push token to `PdfStreamWriter`, if this token have a reserved number, it would be immediately pushed to the token list. On the contrary, if the token doesn't have a reserved number the complexity of the insertion would be `O(n) (n being the amount of token that have already being push)`, this is because we have to check that this token is not duplicated. I was thinking about reducing this complexity by grouping token by type, but this would need some design thoughts, so I left it as it is, so we can discuss it.

One thing to notice, this improvement would not improve all merged document, only those that share similar properties.

I would leave some comment on specific line to discuss some piece of code

### DIFF
--- a/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
@@ -88,5 +88,26 @@
 
             return builder.ToString();
         }
+        
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
+                return true;
+
+            if (!(obj is ArrayToken other))
+                return false;
+
+            if (other.Length != Length)
+                return false;
+
+            for (var index = 0; index < Length; ++index)
+            {
+                if (!Data[index].Equals(other[index]))
+                    return false;
+            }
+
+            return true;
+        }
     }
 }
--- a/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
@@ -92,19 +92,27 @@
         /// <inheritdoc />
         public bool Equals(IToken obj)
         {
-            if (this == obj)
+            if (ReferenceEquals(this, obj))
+            {
                 return true;
+            }
 
             if (!(obj is ArrayToken other))
+            {
                 return false;
+            }
 
             if (other.Length != Length)
+            {
                 return false;
+            }
 
             for (var index = 0; index < Length; ++index)
             {
                 if (!Data[index].Equals(other[index]))
+                {
                     return false;
+                }
             }
 
             return true;
--- a/src/UglyToad.PdfPig.Tokens/BooleanToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/BooleanToken.cs
@@ -30,17 +30,6 @@
         }
 
         /// <inheritdoc />
-        public override bool Equals(object obj)
-        {
-            if (!(obj is BooleanToken other))
-            {
-                return false;
-            }
-
-            return other.Data == Data;
-        }
-        
-        /// <inheritdoc />
         public override int GetHashCode()
         {
             return Data.GetHashCode();
@@ -50,6 +39,20 @@
         public override string ToString()
         {
             return Data.ToString();
+        }
+
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
+                return true;
+
+            if (!(obj is BooleanToken other))
+            {
+                return false;
+            }
+
+            return other.Data == Data;
         }
     }
 }
--- a/src/UglyToad.PdfPig.Tokens/BooleanToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/BooleanToken.cs
@@ -44,8 +44,10 @@
         /// <inheritdoc />
         public bool Equals(IToken obj)
         {
-            if (this == obj)
+            if (ReferenceEquals(this, obj))
+            {
                 return true;
+            }
 
             if (!(obj is BooleanToken other))
             {
--- a/src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
@@ -114,6 +114,26 @@
             return new DictionaryToken(result);
         }
 
+
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
+                return true;
+
+            if (!(obj is DictionaryToken other))
+            {
+                return false;
+            }
+
+            if (Data.Count != other.Data.Count)
+                return false;
+
+            // TODO: Maybe consider using a sorted dictionary?
+            return Data.OrderBy(kvp => kvp.Key)
+           .SequenceEqual(other.Data.OrderBy(kvp => kvp.Key));
+        }
+
         /// <inheritdoc />
         public override string ToString()
         {
--- a/src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
@@ -118,8 +118,10 @@
         /// <inheritdoc />
         public bool Equals(IToken obj)
         {
-            if (this == obj)
+            if (ReferenceEquals(this, obj))
+            {
                 return true;
+            }
 
             if (!(obj is DictionaryToken other))
             {
@@ -127,11 +129,19 @@
             }
 
             if (Data.Count != other.Data.Count)
+            {
                 return false;
+            }
 
-            // TODO: Maybe consider using a sorted dictionary?
-            return Data.OrderBy(kvp => kvp.Key)
-           .SequenceEqual(other.Data.OrderBy(kvp => kvp.Key));
+            foreach (var kvp in other.Data)
+            {
+                if (!Data.TryGetValue(kvp.Key, out var val) || !val.Equals(kvp.Value))
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         /// <inheritdoc />
--- a/src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
@@ -22,8 +22,10 @@
         /// <inheritdoc />
         public bool Equals(IToken obj)
         {
-            if (this == obj)
+            if (ReferenceEquals(this, obj))
+            {
                 return true;
+            }
 
             if (!(obj is InlineImageDataToken other))
             {
@@ -31,14 +33,17 @@
             }
 
             if (Data.Count != other.Data.Count)
+            {
                 return false;
+            }
 
-            // Note: This maybe slow?
-            // Maybe pre calculate some sort of hash and compare that?
+
             for (var index = 0; index < Data.Count; ++index)
             {
                 if (Data[index] != other.Data[index])
+                {
                     return false;
+                }
             }
 
             return true;
--- a/src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
@@ -18,5 +18,30 @@
         {
             Data = data;
         }
+
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
+                return true;
+
+            if (!(obj is InlineImageDataToken other))
+            {
+                return false;
+            }
+
+            if (Data.Count != other.Data.Count)
+                return false;
+
+            // Note: This maybe slow?
+            // Maybe pre calculate some sort of hash and compare that?
+            for (var index = 0; index < Data.Count; ++index)
+            {
+                if (Data[index] != other.Data[index])
+                    return false;
+            }
+
+            return true;
+        }
     }
 }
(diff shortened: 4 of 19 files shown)


src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
```diff
@@ -114,6 +114,26 @@ public DictionaryToken With(string key, IToken value)
...
+
+            if (!(obj is DictionaryToken other))
+            {
+                return false;
+            }
+
+            if (Data.Count != other.Data.Count)
+                return false;
+
+            // TODO: Maybe consider using a sorted dictionary?
```
InusualZ: Is this a good idea? 

As of right know. Every time we compare dictionary we have to sort by their key and then compare. If we use a sorted dictionary, this would only be made once per insertion. But, we would lose the order of insertion, I don't know if that is important enough?

src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
```diff
@@ -18,5 +18,30 @@ public InlineImageDataToken(IReadOnlyList<byte> data)
...
+            {
+                return false;
+            }
+
+            if (Data.Count != other.Data.Count)
+                return false;
+
+            // Note: This maybe slow?
+            // Maybe pre calculate some sort of hash and compare that?
+            for (var index = 0; index < Data.Count; ++index)
```
InusualZ: Like the comment say, should calculate some sort of hash? To avoid full bytes comparison?

src/UglyToad.PdfPig.Tokens/ObjectToken.cs
```diff
@@ -37,6 +37,39 @@ public ObjectToken(long position, IndirectReference number, IToken data)
...
+            if (!(obj is ObjectToken other))
+            {
+                return false;
+            }
+
+            // I don't think we should compare this value...
+            // If this values are the same but this object are not the same (reference)
+            // It means that one of the obj is not from stream or one of the object have been overwritten
+            // So maybe putting an assert instead, would be a good idea?
+            if (Position != other.Position)
```
InusualZ: This make sense? Maybe we should `throw` in this case

src/UglyToad.PdfPig.Tokens/StreamToken.cs
```diff
@@ -29,7 +29,34 @@ public StreamToken(DictionaryToken streamDictionary, IReadOnlyList<byte> data)
...
+            }
+
+            if (!StreamDictionary.Equals(other.StreamDictionary))
+            {
+                return false;
+            }
+
+            // Note: This maybe slow?
+            // Maybe pre calculate some sort of hash and compare that?
+            for (var index = 0; index < Data.Count; ++index)
```
InusualZ: Like the comment say, should calculate some sort of hash? To avoid full bytes comparison?

src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -132,75 +130,59 @@ private class DocumentMerger
...
+            private readonly PdfStreamWriter context = new PdfStreamWriter();
+            private readonly List<IndirectReferenceToken> pagesTokenReferences = new List<IndirectReferenceToken>();
+            private readonly IndirectReferenceToken rootPagesReference;
 
             private decimal currentVersion = DefaultVersion;
-            private MemoryStream memory = new MemoryStream();
-
             private int pageCount = 0;
+            
+            private readonly Dictionary<IndirectReferenceToken, IndirectReferenceToken> referencesFromDocument =
```
InusualZ: This dictionary would keep track of token that are being written from the current document being added to the resultant document. I put like so (globally, class scope), to avoid having to pass it as parameter, cluttering the parameters of `CopyToken`

src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -286,79 +266,111 @@ private ObjectToken CopyPageNode(PageTreeNode pageNode, IndirectReferenceToken p
...
-                    return new ArrayToken(newArray);
+                        var bytes = deepCopy ? new List<byte>(streamToken.Data) : streamToken.Data;
+                        return new StreamToken(properties, bytes);
+                    }
+                    
+                    case ObjectToken _:
+                    {
+                        // This is because, since we don't write token directly to the stream. So can't know the offset.
+                        // The token would be invalid. Although I don't think the copy of an object token would ever happen
+                        throw new NotSupportedException("Copying a Object token is not supported");
```
InusualZ: Is this clear enough?

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public IndirectReferenceToken ReserveNumberToken()
+        { 
+            return new IndirectReferenceToken(new IndirectReference(ReserveNumber(), 0));
+        }
+        
+        public byte[] ToArray()
+        {
+            if (!Stream.CanSeek)
+                throw new NotSupportedException($"{Stream.GetType()} can't seek");
```
InusualZ: Should we let the `Stream` throw this exceptions? As of right know we are missing to check if we can read (`CanRead`)

src/UglyToad.PdfPig.Tokens/InlineImageDataToken.cs
EliotJones: This is fine I think, we don't actually expose this token anywhere that the `Equals` method will be called with any regularity. Since you'd have to do a full scan for hash collisions anyway I'd leave it as is.

src/UglyToad.PdfPig.Tokens/ArrayToken.cs
```diff
@@ -88,5 +88,26 @@ public override string ToString()
 
             return builder.ToString();
         }
+        
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
+                return true;
```
EliotJones: Sorry I realise I never got round to sharing the dotsettings I'm using. In general I'd like to keep to having the curly braces even for single line statements. It's more verbose but also more explicit about what's in scope and stops problems in future maintenance where things might get inadvertently added/removed from a scope. This is a comment across the modified files but I'll just highlight it here.

src/UglyToad.PdfPig.Tokens/ObjectToken.cs
EliotJones: I think maybe we should just remove the position check since you might have a situation where someone is accessing the document structure from different parsed documents using `document.Catalog.Structure` or however it's accessed and they share a producer which means they have identical objects but are in different positions and can be combined. The user can keep `ReferenceEquals` for the non-content aware equality.

src/UglyToad.PdfPig.Tokens/StreamToken.cs
EliotJones: I think this is fine, since the byte array is already in memory it's not going to be worth hashing unless it happens on a tight loop somewhere, I'll profile it to confirm but I don't think it currently is.

src/UglyToad.PdfPig/Writer/PdfMerger.cs
EliotJones: I think this makes sense, as you say I don't think we'd ever be copying an object token because we're creating object tokens by copying and wrapping the non-object token types.

src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -286,79 +266,111 @@ private ObjectToken CopyPageNode(PageTreeNode pageNode, IndirectReferenceToken p
...
-                stream.Write(bytes, 0, bytes.Length);
-                stream.WriteNewLine();
+                    case HexToken hexToken:
+                        return new HexToken(hexToken.Data.ToCharArray());
+                    
+                    case InlineImageDataToken imageDataToken:
+                        return new InlineImageDataToken(imageDataToken.Data);
+                    
+                    case NameToken nameToken:
+                        return NameToken.Create(nameToken.Data);
```
EliotJones: You don't actually need to deep copy the name token afaik, there's an inner cache which ensures only a single name token value for a given string can exist in the AppDomain.

EliotJones: Thinking more about this I don't think we need this deep copy step. Things like NameToken, NumericToken, OperatorToken, etc are cached as far as possible so that we don't allocate too many instances. I think I'm misunderstanding the benefit/fix the cloning brings in this case.

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
EliotJones: I alternate between letting the stream throw it and adding the check myself when I'm feeling eager. I'd add the CanRead check as well and move the check to the constructor in order to fail as early as possible if the stream is invalid, or remove it and let the stream fail when it gets used.

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+            
+            var currentPosition = Stream.Position;
+            Stream.Seek(0, SeekOrigin.Begin);
+
+            var bytes = new byte[Stream.Length];
+
+            // Should we slice the reading into smaller chunks?
+            if (Stream.Read(bytes, 0, bytes.Length) != bytes.Length)
```
EliotJones: One alternative is just to use an intermediary memory stream and let the framework worry about the buffer handling:

```
using (var memory = new MemoryStream())
{
   stream.CopyTo(memory)l
   return memory.ToArray()
}
```

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
```diff
@@ -0,0 +1,166 @@
...
+                throw new Exception("Unable to read all the bytes from stream");
+
+            Stream.Seek(currentPosition, SeekOrigin.Begin);
+
+            return bytes;
+        }
+
+        public void Dispose()
+        {
+            Stream?.Dispose();
```
EliotJones: I generally like to have a flag for this, if the user has provided the stream we probably shouldn't dispose it ourselves. For example they might be writing a streaming web response or streaming some other form of data and they want to continue to use the stream, for example to write an end-of-data marker after using it for our method. On the other hand if we've instantiated our own memory stream we're allowed to dispose it. But since memorystream doesn't require disposing we can probably remove the dispose entirely, or just null our reference.

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
InusualZ: I highly against this.  When we execute this method. The result would be at least 3 buffer of the same size
1. PdfStreamWriter stream
2. MemoryStream memory
3. The copy that we return, when we do `memory.ToArray`

That's without counting the list of token that we haven't already flushed... For simple document, this would not matter much, but for big document, the memory footprint would be huge

src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
InusualZ: I would implement your alternative.

src/UglyToad.PdfPig.Tokens/ObjectToken.cs
InusualZ: That make sense, I didn't think about the `ReferenceEquals`

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
InusualZ: I thought about the possibility of putting the checks in the constructor. But since is possible to use the Stream without ever calling this method, I leave them here. So for that reason I would let the inner stream throw

src/UglyToad.PdfPig.Tokens/BooleanToken.cs
```diff
@@ -51,5 +40,19 @@ public override string ToString()
         {
             return Data.ToString();
         }
+
+        /// <inheritdoc />
+        public bool Equals(IToken obj)
+        {
+            if (this == obj)
```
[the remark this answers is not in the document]
InusualZ: Done deal

src/UglyToad.PdfPig/Writer/PdfMerger.cs
InusualZ: To be honest, I can't remember my reasoning behind implementing the deep copy. I guess it was something related to copy specific things from documents. But I guess I didn't take in consideration, that as of right know. Most of DataToken properties are `readonly` so there is no point.

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
InusualZ: Done deal

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
EliotJones: That's fair enough, I'm happy with it as-is.

PR #170, opened 2020-04-23
added check for bad fonts see https://github.com/apache/pdfbox/blob/6…

busbina:
Was getting an error on parsing a PDF with PDFPig but not on PDFBox. Found this difference that fixed my issue with a bad font causing an "Array dimensions exceeded supported range" exception. 

I can tag this with an issue if you want.

The corresponding PDFBox code is here: https://github.com/apache/pdfbox/blob/61ceca8376f08f23f62178bd4fd97e919a690e43/fontbox/src/main/java/org/apache/fontbox/ttf/HorizontalMetricsTable.java


EliotJones:
Thanks, this looks great, I'll take a look in detail tomorrow.

Many thanks for contributing this fix :)

PR #171, opened 2020-04-27
Set ClipPaths to false if no ParsingOptions given (consistent behaviour)

BobLd:


EliotJones:
Thanks, I missed that

Issue #87, opened 2019-11-24
Handling clipping in ProcessFormXObject


BobLd:
Hi, 

The form (letters and paths) on page 8 in this [document](https://www.researchgate.net/publication/220932988_Performance_Comparison_of_Six_Algorithms_for_Page_Segmentation) are not rendered properly because form clipping is not handled yet (on purpose).

No rush, it is more to put here a reference document with the problem.

I think it would be great to start by hadling clipping for rectangle areas first (then to be extended to generic paths). I'll do a PR with a `Intersect` function for `PdfRectangle` soon.

Thanks!

Issue #161, opened 2020-04-07
Letter positions are off


vadik299:
Hi Eliot,

Could you take a look at this document please?
https://drive.google.com/file/d/1AH7e5ZG4e-G1v256he84bca8txarUwcn/view?usp=sharing

For some reason I'm getting totally wrong letter locations there. E.g. the first letter is at (866,499), which is outside of the page itself. This page is rotated by 90 degrees, so i think Y should be negative. Is it a problem with transformation matrix or something like that?
Thank you!

EliotJones:
Ok, it might be hidden content but sounds like something's wrong here anyway. I'll try to take a look this weekend, thanks.

EliotJones:
Hi @vadik299 , I've had a look and ended up doing some fairly major re-work to how rotation functions so you may need to update code which handles rotated documents.

Now the page coordinates, width and height are corrected for rotation. So the width and height of the page now match the actual size after the rotation is applied. In addition letter positions are no longer negative, the coordinate system is updated to account for the rotation so letters now have values that correspond to the correct page location.

For example (ignore the red boxes, that's the default word extractor being bugged), the output letter boxes without any additional calculations to account for rotation are now correct:
![image](https://user-images.githubusercontent.com/7862346/79644278-d154cc00-819f-11ea-9130-a7326187512c.png)


Please give it a go and let me know if you run into problems with this new approach or need any more details.

vadik299:
Hi Eliot,

Thank you so much for the fix! I adjusted logic for the page rotation accordingly and it works well!

I'm using function page.ExperimentalAccess.GetPointSize() to get text height, because Letter.FontSize is not very reliable. (I think you recommended this in the past). Now I noticed that it stopped working for some documents where it was working before. So I had to add a work-around, to fallback to FontSize when GetPointSize returns zero. If you think these are worth another look I can find a few examples where they don't return same values (for some documents one works well, for other documents the other one).

EliotJones:
Glad to hear this worked for you.

I have just pushed up an additional change to the point size calculation, for letters which had some rotation applied the value was 0 previously. The value will still be incorrect where the rotation is not through some 90 degree angle.

Please try it out and let me know how you get on.

BobLd:
Do you think we could use the point size in bounding box by default when height was 0 previously?

vadik299:
Thank you Eliot, 
GetPointSize() is working perfectly now!

PR #179, opened 2020-06-01
Improve ContentStreamProcessor

BobLd:
- Fix the multiplication order of the new `CurrentTransformationMatrix` in `ProcessFormXObject()`.
This is in line with the pdf documentation: _b) Concatenates the matrix from the form dictionary’s Matrix entry with the current transformation matrix (CTM)_ -> Fix issue #87 , which in fact was not related to clipping
- better handling of page rotation by updating initial `TransformationMatrix`, linked to #161. Unchanged behavior for `Letter`, but now all elements are handled, e.g. `PdfPaths`


EliotJones:
All visual tests look good (output of rotated text bounding boxes can be seen in `./src\UglyToad.PdfPig.Tests\bin\Debug\netcoreapp2.0\Images` after a test run) and the performance is slightly improved over the current master performance 👍 merging

Issue #180, opened 2020-06-09
PDFTextStripper class


ivanicin:
First to say that I was (and I am still) very happy to see that there is a port of PDFBox to .NET.

I understand that it is work in progress, but I would like to stress the need to port the `PDFTextStripper` class which doesn't appear to be ported at the moment. For me and probably for many other developers it is the only need why I use PDFBox at all.

Hopefully it will be implemented in the future. Looking forward on it. If it is possible keep this issue open until it is implemented so that all people interested in this feature can get easily notified on that change.

BobLd:
Hi @ivanicin, thanks for the feedback! Can you give example on how you'd use `PDFTextStripper`, is it to extract text? If it's the case, what would be the difference with PdfPig's `ContentOrderTextExtractor`?

ivanicin:
> Hi @ivanicin, thanks for the feedback! Can you give example on how you'd use `PDFTextStripper`, is it to extract text? If it's the case, what would be the difference with PdfPig's `ContentOrderTextExtractor`?

This seems to be it though I couldn't make it work but that's another issue that I'll file separately.

Currently this is in pre-release nuget, so not sure if you close the bug until it is in the regular release or as soon as you get some sort of solution, so I'll leave to owners to decide on whether to close the issue immediately or not.

ivanicin:
Also I am not sure if this should be the same or new issue, but when I try to parse this file: https://s3-us-west-2.amazonaws.com/pressbooks-samplefiles/LewisTheme/The-Problems-of-Philosophy-LewisTheme.pdf it appears that results are far from `PDFTextStripper` quality - at first glance I can't see any problems with `PDFTextStripper`, while there are significant problems with `ContentOrderTextExtractor`, like missing spaces where expected or new lines where not expected or doubling some words, they appear on many places in this text so it is quite easy to spot.

And just to note that this document isn't particularly complex so probably there are even worse problems on more complex documents.

So I would leave this issue open until you can get the similar quality on this sample document of `ContentOrderTextExtractor`  when compared to `PDFTextStripper` (which should be feasible as it is part of PDFBox that you try to port), or we can close this issue and I can file a new one if that is preferred.

BobLd:
for reference on the duplicate letters, this is how `PDFTextStripper` seems to handle it:

[...]

PR #185, opened 2020-06-25
Implement suppressDuplicateOverlappingText

BobLd:
needed for #180

### DIFF
--- a/src/UglyToad.PdfPig/Content/IPageFactory.cs
+++ b/src/UglyToad.PdfPig/Content/IPageFactory.cs
@@ -4,6 +4,6 @@
 
     internal interface IPageFactory
     {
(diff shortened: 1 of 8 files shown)


EliotJones:
Thanks for looking into this. I'm going to be annoying though because I think the filtering is at the wrong level. It seems more like a post-processing step that should be done by DLA tools such as the textordercontentextractor, perhaps a util inside DLA can take in the Enumerable of letters and returned the duplicate filtered Enumerable.

Doing it at the content stream level risks removing valid data and will have a large performance impact so I'd much rather avoid doing it here. In general the idea with the content stream processor is to naively return all content, irrespective of its meaning, and then additional tools can post-process that content to arrive at the preferred content.

BobLd:
Didn't think about it but makes complete sense! Will push a stand-alone util to do this (maybe later on we'll need to add IPre/IPostProcessing interfaces)

BobLd:
see #186

PR #186, opened 2020-06-25
Implement DuplicateOverlappingTextProcessor

BobLd:
Implement `DuplicateOverlappingTextProcessor`
The downside compared to #185 is that the duplicated letter's font details cannot be updated with `Bold` (`Letter` is internal).

Any comment welcome

### DIFF
--- a/src/UglyToad.PdfPig.DocumentLayoutAnalysis/DuplicateOverlappingTextProcessor.cs
+++ b/src/UglyToad.PdfPig.DocumentLayoutAnalysis/DuplicateOverlappingTextProcessor.cs
@@ -3,15 +3,16 @@
     using System.Collections.Generic;
     using System.Linq;
     using UglyToad.PdfPig.Content;
+    using UglyToad.PdfPig.PdfFonts;
 
     /// <summary>
-    /// Checks if each letter is a duplicate and overlaps any other letter and remove the duplicate. Duplicate overlapping letters might be used to simulate bold.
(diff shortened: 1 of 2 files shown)


src/UglyToad.PdfPig.DocumentLayoutAnalysis/DuplicateOverlappingTextProcessor.cs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    cleanLetters.Add(letter);
+                }
+                else
+                {
+                    // TODO: update font details to bold
+                    // TODO: need to update the bounding box
+                    // TODO: need to update bottom left/right
+                    // TODO: need to update width
+                    // update textSequence?
```
EliotJones, line 66: I think the approach for these would be to create a new letter in the result list, so we remove the duplicate and use the Letter constructor to create a new one with the right details (though I don't recall if letter constructor itself is internal). Would that work do you think?

BobLd: This is what I tried, but because the `Letter` constructor is internal, I cannot access it from the DLA lib. So I would need to make the constructor `public`... do you have another idea?

EliotJones: Yes sorry I forgot the constructor was internal, making it public (as you've just committed) is good

Issue #27, opened 2019-01-01
Enable loading a document from a PdfDocument into a PdfDocumentBuilder


EliotJones:
In order for the library to be useful it must support editing as well as creating and reading. For this we need a way to read an existing document and convert it to a document builder.

davidpene:
This would make an awesome addition 💯 
We're currently using PDFSharpCore but it doesn't have quite as nice of an interface to use and we see a lot of bugs from it.

Numpsy:
Sort of related to this, plus possibly related to merging and/or perhaps #193:

I have a use case (currently done with PDF Clown) for updating the document info dictionary in an existing PDF, but not making any other changes (that is both modifying the 'known'/default properties and adding/removing/modifying custom ones).

It's been a while since I looked at the PDFs that PDFClown produces, but I think the simple cases are handled by appending a new info dictionary to the end of the file, then updating the file to reference that (such that the previous dictionary contents are left in the file as they were). (actually, I think i raised a bug some time back about PdfPig finding the wrong dictionary instance in some of those PDFs, which was fixed).

So, is there a simple way to do that with the current PdfPig? (i mention merging etc as I wonder if you could merge in the info dictionary from a new document? Not sure if that's a bit round about though?

EliotJones:
Hi @Numpsy unfortunately I don't think there's any way to do this with the existing PdfPig code currently. I made a start on what I think needs to be done in https://github.com/UglyToad/PdfPig/tree/edit-existing-documents but I never seem to have time anymore for deep work on PdfPig :(

The aim (sorry for the terrible quality diagram) is to change PdfDocumentBuilder to be an API wrapper that directly modifies an in-memory tree of the underlying PDF objects:

![image](https://user-images.githubusercontent.com/7862346/89731141-ae933f80-da3c-11ea-992b-041354790654.png)

The tree (on the right) is similar to what you'd see if you inspect a PDF document in the iTextRups desktop application, in the case of a new PDF document the tree is empty (except for Catalog/Trailer/InfoDictionary) and operations against the PdfDocumentBuilder add pages, fonts, etc to the tree and the user can also access the tree direcly. For an existing document the tree is fully populated and the PdfDocumentBuilder is populated with the corresponding pages and fonts entries and accessed in the same way as for new documents.

[...]

PR #193, opened 2020-07-27
New API for copying token from another document

InusualZ:
**Description WIP**

### What need to be done
- [x] Use this API in the PdfMerger
- [x] Deduplicating fonts (FontCopier)
- [ ] Font Merging (At least for TrueTypeFont)
- [ ] Detecting duplicated page content stream. Replacing the streams with an XObject

### Later
- [ ] Multi Threading (Performance Boost)

## **Comments are encourage**


src/UglyToad.PdfPig.Fonts/TrueType/Tables/CMapSubTables/Format4CMapTable.cs
```diff
@@ -56,6 +57,58 @@ public int CharacterCodeToGlyphIndex(int characterCode)
...
+                    charactersCode.AddRange(chars);
+                }
+            }
+
+            if (minCharacterStart < 0xf000)
+            {
+                return charactersCode.ToArray();
+            }
+
+            if (minCharacterStart < 0xf100)
```
InusualZ, line 96: I gathered this masks from [this](https://github.com/UglyToad/PdfPig/blob/dc55fd80113f8145b4f00f4afdb7fa808dc4a9fa/src/UglyToad.PdfPig.Fonts/TrueType/Tables/CMapTable.cs#L72-L96), but I don't know if there is a pattern that I should follow

src/UglyToad.PdfPig.Fonts/TrueType/Tables/CMapSubTables/Format4CMapTable.cs
```diff
@@ -56,6 +57,58 @@ public int CharacterCodeToGlyphIndex(int characterCode)
...
+                if (segment.IdRangeOffset == 0)
+                {
+                    var chars = Enumerable.Range(segment.StartCode, characterCountInSegment);
+                    charactersCode.AddRange(chars);
+                }
+                else
+                {
+                    var chars = Enumerable.Range(0, characterCountInSegment).Where(i =>
+                    {
+                        var offset = segment.IdRangeOffset / 2 + i;
```
InusualZ, line 82: This filter (`Where` Clause) it's unnecessary. But, since I'm really no expert in TrueType fonts it gives me a little more security having it there.

src/UglyToad.PdfPig/Writer/Copier/Font/FontObject.cs
```diff
@@ -0,0 +1,174 @@
...
+    using Filters;
+    using PdfPig;
+    using PdfPig.Fonts.TrueType;
+    using PdfPig.Fonts.TrueType.Parser;
+    using PdfPig.Fonts.Type1;
+    using PdfPig.Fonts.Type1.Parser;
+    using Tokens;
+    using Util.JetBrains.Annotations;
+
+    internal class FontObject
```
InusualZ, line 16: I'm not a fan of this data struct. Suggestion are welcome.

src/UglyToad.PdfPig/Writer/Copier/TokenHelper.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace UglyToad.PdfPig.Writer.Copier
+{
+    using System;
+    using System.IO;
+    using Core;
+    using Tokens;
+
+    internal static class TokenHelper
```
InusualZ, line 8: I thought about using the `DirectObjectFinder` helper class that we have, but I think this method offer more flexibility in term of what can be a token scanner

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
```diff
@@ -151,21 +181,6 @@ private IndirectReferenceToken AddToken(IToken token, int reservedNumber)
             return referenceToken;
         }
 
-        private IndirectReferenceToken FindToken(IToken token)
```
InusualZ, line 154: I removed this method, because It was causing a huge performance drop with big file. I need to come up with a better idea.

EliotJones (commented):
Sorry it's taken so long to get round to a proper review here!

It's a little difficult to review and I'll need to clone it locally to have a proper look. I think however we might be better merging the merging API with the approach described in this comment https://github.com/UglyToad/PdfPig/issues/27#issuecomment-671041414 for editing existing documents. Though we need a lot of the code from this PR for fonts, and it makes sense to use it for intelligent merging of font objects, both in building and merging documents.

The main task is to create a mutable tree describing the edited/merged document. So on the first document loaded into the merger/document builder populates the initial object tree:

```
// Naming is difficult
public class MutablePdfObjectTree
{
    private readonly Dictionary<IndirectReference, ObjectToken> referenceMap = new...;

    public DictionaryToken Root { get; private set; }

    public PagesTree Pages { get; private set; }

    // Lots more stuff   
}
```

The on the addition of subsequent documents we can decide how to merge trailers, info dicitionaries, page trees, fonts, etc. 

If you want to finish off this PR though as the next step that's fine. There are a few comments on some style tweaks, etc. I'm hoping to look into #199 with respect to merging soon too so maybe this will also fix that.

src/UglyToad.PdfPig/Writer/Copier/Font/FontObject.cs
```diff
@@ -0,0 +1,174 @@
...
+                !fontDescriptor.TryGet(NameToken.FontFile3, out tokenObj))
+            {
+                // This would mean that the font is not embedded
+                return (null, null);
+            }
+
+            // TODO: Check if we always should get a StreamToken
+            var fontFileStream = TokenHelper.GetTokenAs<StreamToken>(tokenObj, lookupFunc);
+
+            var bytes = fontFileStream.Decode(DefaultFilterProvider.Instance).ToArray();
```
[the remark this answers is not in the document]
InusualZ: Because. `Decode` returns a `IReadOnlyList` and most of the method that I'm calling with the bytes take a `byte[]`

src/UglyToad.PdfPig/Writer/PdfStreamWriter.cs
plaisted: I wrote something similar for a different library. During writing of tokens it would create a hash (FNV but not really important) of the written contents (excluding the object header and footer) and store it in a dictionary. Subsequent writes would check to see if the hash had been written and use the existing ref if it had. The hashing was so quick that it didn't affect performance noticeably for large documents.

InusualZ:
I'm closing this since this would never be finished

Issue #189, opened 2020-07-16
Indirect reference to an undefined object behavior


Poltuu:
I stumbled upon an error on a specific pdf using `document.GetPages()`.

The error occurs because the fonts node
```
/Font <<
/F0 8 0 R
/F1 9 0 R
/F2 10 0 R
/F3 0 0 R
>>
```

makes reference to an undefined font `/F3 0 0 R`.
This ends up failing  with the following stack
```
   at UglyToad.PdfPig.Tokenization.Scanner.PdfTokenScanner.Get(IndirectReference reference)
   at UglyToad.PdfPig.Parser.Parts.DirectObjectFinder.Get[T](IToken token, IPdfTokenScanner scanner)
   at UglyToad.PdfPig.Content.ResourceStore.LoadFontDictionary(DictionaryToken fontDictionary)
   at UglyToad.PdfPig.Content.ResourceStore.LoadResourceDictionary(DictionaryToken resourceDictionary)
   at UglyToad.PdfPig.Parser.PageFactory.Create(Int32 number, DictionaryToken dictionary, PageTreeMembers pageTreeMembers, Boolean clipPaths)
   at UglyToad.PdfPig.Content.Pages.GetPage(Int32 pageNumber, Boolean clipPaths)
   at UglyToad.PdfPig.PdfDocument.GetPage(Int32 pageNumber)
   at UglyToad.PdfPig.PdfDocument.<GetPages>d__32.MoveNext()
```

I looked at the doc, and it seems that the official behavior would be to make a reference to null, cf https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf

I'm more than willing to propose a PR if that works for you.

EliotJones:
Hi, thanks for your issue report. If you have a PR that would be fantastic, otherwise I can try to fix it this weekend. Feel free to open a PR if you have time :)

Poltuu:
I'm unable to push a branch, you may need to add me as a contributor ?

InusualZ:
> 
> 
> I'm unable to push a branch (github rejects it), you may need to add me as a contributor ?

You should send a Pull Request. So we can review your changes first

Poltuu:
:suspect: but how would I send a pull request if i'm unable to push my branch into the repository 🤔

EliotJones:
[...]

PR #197, opened 2020-08-17
FIX : undefined references is a valid use case.

Poltuu:
Fixes https://github.com/UglyToad/PdfPig/issues/189
I tried to mitigate the breaking change by keep on throwing in most uses of the change method.


EliotJones:
Thanks for this fix

PR #200, opened 2020-08-19
FIX Error when getting Encoding from an IndirectReferenceToken referencing a Name Token

everharder:
if baseEncodingObject is a IndirectReferenceToken to a NameToken the previous implementation tried to resolve it as a DictionaryToken, which resultet in an PdfFormatException.


EliotJones:
Thanks for this :)

everharder:
I have to thank you for your awesome library. Keep up the good work!

Issue #206, opened 2020-08-28
Finding out Letter background color


Poltuu:
I stumbled upon white text written on white background (so basically hidden), and I wondered how easy it would be to find out the background color of a `Letter` ? 

I'm guessing it's basically impossible for certain, but just asking in case there is something viable.

EliotJones:
Hi @Poltuu, the default is for a white background (I don't think there's a way to change the whole page background color) but the things that could affect this are any PdfPaths with a fill color drawn in the region, or any images placed behind the letter. The API provides access to the paths through `page.ExperimentalAccess.GetPaths()` but they're a pain to work with, and images complicate it even further.

Poltuu:
Thanks, that's a good first step for me.

BobLd:
To add to what @EliotJones said, maybe you can start with the approach below to check for background:
```csharp
using UglyToad.PdfPig.Geometry;
// [...]

var parsingOptions = new ParsingOptions() { ClipPaths = true };
using (PdfDocument document = PdfDocument.Open(pdfPath, parsingOptions))
{
	var page = document.GetPage(1);

	var images = page.GetImages().ToList();

	// get letters on top of images
	foreach (var image in images)
	{
		var onTop = page.Letters.Where(l => image.Bounds.Contains(l.GlyphRectangle)).ToList();

		// need to find a way to know the image color where the letters are
		// [...]
	}

	// get filled paths
	var filledPaths = page.ExperimentalAccess.Paths.Where(p => p.IsFilled).ToList();

	// get letters on top of filled path
	foreach (var filledPath in filledPaths)
	{
		var fillColor = filledPath.FillColor.ToRGBValues();
		// need to do some checks on the fill color here, e.g. reject if filled with white
		// [...]

		// we approximate the PdfPath to its axis aligned bounding box (rectangle)
		var bbox = filledPath.GetBoundingRectangle(); 

		if (bbox.HasValue)
		{
			var onTop = page.Letters.Where(l => bbox.Value.Contains(l.GlyphRectangle)).ToList();

			// need to do some check between letter's color and pdfpath fill color here
			// [...]
		}
	}
}
```

To compare colors, you could simply use equality, but in more complex cases you could use a distance measure. If two colors are too close, they are considered the same. Easiest way to do so is to use the euclidian distance, but it has its drawbacks. I did some tests on color distances [here](https://github.com/BobLd/PdfPig/tree/colour-distance). It is a more complicated topic than it seems.

[...]

PR #207, opened 2020-08-29
Adding Pdfpath/PdfSubpath extensions

BobLd:
- Adding `Contains()` methods to `PdfPath` and `PdfSubpath` for easier handling (leverages the Clipper library)
- Adding tests for PdfPath.Contains(PdfRectangle) - see image below. Words ending in "_in" are inside pdfpaths, those ending in "_out" are outside. Handles holes in shapes. Will be useful for #206 

NB: more tests are needed
image


BobLd:
I'm not completely happy with the method name `Contains()` for the `PdfPath` as it already has this method as part of `List<PdfSubpath>`. Any feedback welcome

InusualZ:
LGTM.

I couldn't think of a better name for `Cotains()`, aside from `ContainsPath` or `IncludePath`?

PR #210, opened 2020-09-10
ADD : name on type 3 fonts

Poltuu:
I stumbled upon a type 3 font with a name, which seems required in PDF 1.0, according to the spec https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf 

The current code does not take it into account, this PR fixes that.

### DIFF
--- a/src/UglyToad.PdfPig/PdfFonts/Parser/Handlers/Type3FontHandler.cs
+++ b/src/UglyToad.PdfPig/PdfFonts/Parser/Handlers/Type3FontHandler.cs
@@ -58,14 +58,9 @@
 
         private NameToken GetFontName(DictionaryToken dictionary)
         {
-            if (!dictionary.TryGet(NameToken.Name, out var fontName))
+            if (dictionary.TryGet(NameToken.Name, scanner, out NameToken fontName))
             {
-                return NameToken.Type3;
-            }
-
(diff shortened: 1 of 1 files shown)


src/UglyToad.PdfPig/PdfFonts/Parser/Handlers/Type3FontHandler.cs
```diff
@@ -49,11 +49,28 @@ public IFont Generate(DictionaryToken dictionary)
...
+
+            var name = GetFontName(dictionary);
+
+            return new Type3Font(name, boundingBox, fontMatrix, encoding, firstCharacter,
                 lastCharacter, widths, toUnicodeCMap);
         }
 
+        private NameToken GetFontName(DictionaryToken dictionary)
+        {
+            if (!dictionary.TryGet(NameToken.Name, out var fontName))
```
EliotJones, line 61: This might be better as:
```suggestion
            if (!dictionary.TryGet(NameToken.Name, scanner, out NameToken fontName))
```

If I remember correctly the overload which takes a `IPdfTokenScanner` which should be available here will correctly find the name if it lives in a different object, which sometimes happens, otherwise the change looks good to me :+1:

PR #224, opened 2020-10-03
Writer: New API to copy token

InusualZ:
I'm introducing a new API that would allow us to track any type of token while is being copied to a stream, so when a similar token come again, we can decide if want to just use the already written token or the new one.

This API would allow to have separated code (`IObjectCopier`) for deduplicating any type of resource, while not penalizing performance, and thus if any `IObjectCopier` it's causing some performance regression that the user don't want, the user could decide not to add it to the main `IObjectCopier` and the resultant pdf would still be valid

This PR is the first part of #193, for easier review.

All the related test pass normally, except **PdfMergerTests::ObjectCountLower**, but it's expected since the writer is not trying to remove duplicated token and I did not include any `IObjectCopier` that tries to minimize the amount of duplicated token.

PD: While I was preparing this PR, I thought about calling the interface `ITokenCopier` instead of `IObjectCopier`, thoughts?

### DIFF
--- a/src/UglyToad.PdfPig/Writer/Copier/IObjectCopier.cs
+++ b/src/UglyToad.PdfPig/Writer/Copier/IObjectCopier.cs
@@ -1,0 +1,24 @@
+﻿namespace UglyToad.PdfPig.Writer.Copier
+{
+    using System;
+    using Tokens;
+
+    /// <summary>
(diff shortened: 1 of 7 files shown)


EliotJones (approved):
Looks good to me, I think

PR #228, opened 2020-10-09
Switch drawing to independent `IOperationContext` and access to font's bytes stream

BobLd:
**Major changes:**
- switch drawing to independent `IOperationContext` with `SystemDrawingProcessorV2` (related to my point about investigating further using a different `IOperationContext` using directly `System.Drawing`'s `Matrix` and `GraphicsPath`)
- allow access to font's bytes stream (related to my point about exposing more precise font properties)
- expose font family in font details
- start implementing and `PdfShading` and `PdfFunction`

**NB:** the documentation in not done properly on purpose, will be dealt with when it works

**Details:**
- start using `CurrentSystemGraphicsState`
- add `SystemDrawingColorSpaceContext`
- remove BaseDrawingProcessor and SystemDrawingProcessor
- remove reference to `TextMatrices` in `IOperationContext`
- externalise all text operations (`BeginText`, `EndText`, `MoveToNextLineWithOffset`, `SetTextMatrix`) to `ContentStreamProcessor`
- the following are made public:
   - IResourceStore, 
   - MediaBox, 
   - PdfVector, 
   - ResourceColorSpace, 
   - InlineImageBuilder, 
   - XObjectContentRecord, XObjectType, XObjectFactory, 
   - IPageContentParser, 
   - CharacterBoundingBox, 
   - IFont, 
   - IVerticalWritingSupported
- make PdfExtensions.TryGet<T>(...) public
- add `IFont.TryGetDecodedFontBytes(IPdfTokenScanner pdfTokenScanner, IFilterProvider filterProvider, out IReadOnlyList<byte> bytes)` and implement it in `Type0Font`, `TrueTypeSimpleFont`, `TrueTypeStandard14FallbackSimpleFont`, `Type1FontSimple` (still need to check `Type1Standard14Font` and `Type3Font`)
- add `IInputBytes.Bytes` for direct access to the bytes. implement it in `ByteArrayInputBytes` (still need to do it in `StreamInputBytes`)
- add `TrueTypeDataBytes.InputBytes`
- make `ResourceStore`, `PdfTokenScanner`, `FilterProvider` and `PageContentParser` accessible through `page.Experimental`
- add tests

**Known issues:**
- problem with TextMatrix

**Useful resources:**
- https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaDrawingGuide/Transforms/Transforms.html
- https://web.archive.org/web/20170313145219/https://blog.andreloker.de/post/2008/07/03/Load-a-font-from-disk-stream-or-byte-array.aspx


src/UglyToad.PdfPig.Core/StreamInputBytes.cs
```diff
@@ -14,6 +15,11 @@ public class StreamInputBytes : IInputBytes
 
         private bool isAtEnd;
 
+        /// <summary>
+        /// Bytes
+        /// </summary>
+        public IReadOnlyList<byte> Bytes => throw new NotImplementedException();
```
EliotJones, line 21: I guess this might have to be a lazily evaluated property that reads the entire stream into memory for supported stream types `File`, `Memory` and throws otherwise (`HttpContent`)?

src/UglyToad.PdfPig.Fonts/TrueType/TrueTypeFont.cs
```diff
@@ -11,6 +11,11 @@
     /// </summary>
     public class TrueTypeFont
     {
+        /// <summary>
+        /// FontFileBytes
+        /// </summary>
+        public IReadOnlyList<byte> FontFileBytes { get; }
```
EliotJones, line 17: Presumably you need these bytes to pass to the corresponding graphics system, e.g. System.Drawing. I'd avoided adding them to prevent memory bloat in most client code scenarios. Could we instead store some way of accessing the bytes separately to the font object and only access them on demand, e.g. `PdfDocument.GetRawFontBytes(font)` or similar so we don't hold potentially large files in memory?

src/UglyToad.PdfPig.SystemDrawing/SystemDrawingProcessorV2.cs
```diff
@@ -0,0 +1,1287 @@
...
+    using UglyToad.PdfPig.Core;
+    using UglyToad.PdfPig.Graphics;
+    using UglyToad.PdfPig.Graphics.Colors;
+    using UglyToad.PdfPig.Graphics.Core;
+    using UglyToad.PdfPig.Graphics.Shading;
+    using UglyToad.PdfPig.PdfFonts;
+    using UglyToad.PdfPig.Tokens;
+    using UglyToad.PdfPig.XObjects;
+
+    public class SystemDrawingProcessorV2 : IOperationContext, IDrawingProcessor
```
EliotJones, line 23: I'm assuming this is still work in progress, but the `V2` makes me unhappy, but it'll do for now :D

src/UglyToad.PdfPig/Function/PdfFunction.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        /// <summary>
+        /// Clip the given input value to the given range.
+        /// </summary>
+        /// <param name="x">x the input value</param>
+        /// <param name="rangeMin">rangeMin the min value of the range</param>
+        /// <param name="rangeMax">rangeMax the max value of the range</param>
+        /// <returns>the clipped value</returns>
+        protected float clipToRange(float x, float rangeMin, float rangeMax)
```
EliotJones, line 152: I'd prefer to avoid Java style naming even for private/protected methods, so `ClipToRange` and `Interpolate` would be more in keeping with the existing code style.

src/UglyToad.PdfPig/Function/PdfFunction.cs
```diff
@@ -0,0 +1,224 @@
...
+        /// The function dictionary.
+        /// </summary>
+        public DictionaryToken FunctionDictionary { get; }
+
+        /// <summary>
+        /// (Required) An array of 2 × m numbers, where m shall be the number of input values. For each i from 0 to m − 1,
+        /// Domain2i shall be less than or equal to Domain2i+1, and the ith input value, xi, shall lie in the interval
+        /// Domain2i ≤ xi ≤ Domain2i+1. Input values outside the declared domain shall be clipped to the nearest boundary value.
+        /// </summary>
+        public ArrayToken Domain { get; }
```
EliotJones, line 23: Do we want to convert to either a `float[]/double[]/decimal[]/NumericToken[]` at the point this type is created rather than exposing the raw `ArrayToken` which needs casting and error handling every time?

src/UglyToad.PdfPig/Function/PdfFunctionType0.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                retval = new PdfRange(Encode, paramNum);
+            }
+            return retval;
+        }
+
+        /// <inheritdoc/>
+        public override float[] Eval(float[] input)
+        {
+            throw new NotImplementedException("Eval");
```
EliotJones, line 137: Do we have PDFBox code we can use to implement these?

src/UglyToad.PdfPig/Graphics/InlineImageBuilder.cs
```diff
@@ -11,12 +11,30 @@
...
+        /// <summary>
+        /// CreateInlineImage
+        /// </summary>
+        /// <param name="transformationMatrix"></param>
+        /// <param name="filterProvider"></param>
+        /// <param name="tokenScanner"></param>
+        /// <param name="defaultRenderingIntent"></param>
+        /// <param name="resourceStore"></param>
+        /// <returns></returns>
         public InlineImage CreateInlineImage(TransformationMatrix transformationMatrix, IFilterProvider filterProvider,
```
EliotJones, line 38: We could probably keep this method `internal` unless it's used by the other projects, it's not really something the client code will ever invoke and just makes the API noisy?

src/UglyToad.PdfPig.SystemDrawing/SystemDrawingProcessorV2.cs
BobLd: yes indeed, it's wip so no need to worry 😄

src/UglyToad.PdfPig/Function/PdfFunction.cs
BobLd: agreed

src/UglyToad.PdfPig/Function/PdfFunctionType0.cs
BobLd: yes, they are available here: [pdfbox/pdmodel/common/function/](https://github.com/apache/pdfbox/tree/trunk/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/common/function)

src/UglyToad.PdfPig/Graphics/InlineImageBuilder.cs
BobLd: agreed with you to not make the API noisy 😄 . I'll need to check if we really need it `public`

src/UglyToad.PdfPig.Fonts/TrueType/TrueTypeFont.cs
BobLd: Yes we need them to draw the font and yes it would be amazing to have `PdfDocument.GetRawFontBytes(font)`. I'm really not a fan of the what I did.

EliotJones:
Hi @BobLd I'm guessing you're as busy as I am now and don't really have time to finish this one out, I'm going to close it for housekeeping purposes. Feel free to re-open if you ever want to revisit.

PR #229, opened 2020-10-12
Fix Letter's PointSize computation

BobLd:
- fix `Letter`'s `PointSize` computation by applying the transform to a rectangle of height `fontSize`
- add test with rotated letters

Issue was related to https://stackoverflow.com/questions/48010235/pdf-specification-get-font-size-in-points


InusualZ:
Since you are fixing the calculation, you may want to fix the description too [here](https://github.com/UglyToad/PdfPig/blob/7d166131ad25454cc7899b1f0ee943b7f7815179/src/UglyToad.PdfPig/Content/Letter.cs#L71)

EliotJones (approved):
I assume this produces the correct results, thanks for fixing it :)

PR #245, opened 2020-12-09
PdfMerger: Allow to select the pages when merging documents

InusualZ:
This PR allow to pass a page selection to the PdfMerger API, that way we can extract or split a pdf however you want.

**There is currently one drawback: Your page selection must be in order.** This is very important because, if you try to select pages out of order, the resultant pdf would have less pages than the expected amount.

### DIFF
--- a/src/UglyToad.PdfPig/Writer/PdfMerger.cs
+++ b/src/UglyToad.PdfPig/Writer/PdfMerger.cs
@@ -16,6 +16,7 @@
     using Tokenization.Scanner;
     using Tokens;
     using Exceptions;
+    using System.Linq;
 
     /// <summary>
     /// Merges PDF documents into each other.
@@ -29,7 +30,7 @@
         /// <summary>
         /// Merge two PDF documents together with the pages from <paramref name="file1"/> followed by <paramref name="file2"/>.
         /// </summary>
-        public static byte[] Merge(string file1, string file2)
+        public static byte[] Merge(string file1, string file2, IReadOnlyCollection<int> file1Selection = null, IReadOnlyCollection<int> file2Selection = null)
         {
             if (file1 == null)
             {
@@ -45,7 +46,7 @@
             {
                 File.ReadAllBytes(file1),
                 File.ReadAllBytes(file2)
-            });
+            }, new [] { file1Selection, file2Selection });
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -214,30 +263,41 @@ public void Close()
...
                         var count = 0;
-                        (newEntry, count) = CopyPagesTree(pageNode, currentNodeReference, tokenScanner);
+                        (newEntry, count) = CopyPagesTree(pageNode, currentNodeReference, tokenScanner, canCopyPage);
                         nodeCount += count;
                     }
                     else
                     {
-                        newEntry = CopyPageNode(pageNode, currentNodeReference, tokenScanner);
-                        ++nodeCount;
+                        if (canCopyPage())
```
Poltuu, line 285: Is`pageNode.PageNumber` reliable here ? If so, I would have expected `canCopyPage` to be an `ISet<int> selectedPages`. This could also lead, by changing the loop logic line 225/274 https://github.com/UglyToad/PdfPig/pull/245/files?diff=unified&w=1#diff-494177227c2134c490310f8e8141d5033956b4872afedaf2a9cfc3d30fe27816R274 to allow pages reordering.

InusualZ: You are totally right, I should use an `ISet`

The problem with copying pages out of order arise when we take in considerations the resources that those pages share with each others. 

Example

This is my main document
- Root (Page Tree)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 1)
      - Resources
        - Image (F3) (User Icon)
    - Page (Page No. 2)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Another Logo)
      - Font (F2) ( Another Custom Font)
    - Page (Page No. 3)
      - Resources
        - Image (F3) (User Icon)


Now, let's try to merge the selection of pages (1, 3, 2) in that order:

- Root (Page Tree)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 1)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Another Logo)
      - Font (F2) ( Another Custom Font)
    - Page (Page No. 3)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 2)
      - Resources
        - Image (F3) (User Icon)

As you can see we end up with duplicated resources and resources name could collide. This is no problem with small files. But file with many pages (>1000) that share many resources, we would bloat the file and the computer resources.

My idea for solving this was to copy pages from page to tree (bottom to top) instead of tree to page (top to bottom). That way we can walk up the resources, making sure none of them collide and reusing the `IndirectReference` when the resource is reuse. But, this approach is a little more difficult to implement.

InusualZ:
Superseded by #248

PR #246, opened 2020-12-09
Fix #199

### LINKED ISSUE #199: Stackoverflow on merge
ca.pdf is a largish 40 page pdf which I am trying to test merge with a test pdf that is only one page. The single page pdf works fine if I try to merge it with itself, but the 40 page on gives a stackoverflow error.  

var x = PdfMerger.Merge(@"C:\Content\ca.pdf", @"C:\Content\World_Wide_Corp_lorem.pdf");

▶ | $exception | {System.StackOverflowException: Exception of type 'System.StackOverflowException' was thrown.} | System.StackOverflowException

InusualZ:
In large file brute forcing the search of a specific token can be quite difficult. Detecting duplicated token can be made in a more performance way.

This theory was tested on a production program


EliotJones (approved):
LGTM

PR #248, opened 2020-12-12
[Improved] PdfMerger: Allow to select the pages when merging documents

### LINKED ISSUE #244: Pdf splitter
Hi,

I'm contemplating developping a basic pdf splitter functionality based on this library.
I was wondering if you believe it sounds manageable with the current apis (public or internal), and/or if you'd be interested to add this to the library.

At first glance, it seems to me like it's a matter of `PdfPageBuilder` meet `PageContent`, but of course, nothing is that easy with pdfs.

InusualZ:
Supersede #245 

This PR offers a much better option and flexible approach to allow a page selection when merging documents. Also, I address the drawback of the other PR which was related to copying pages out of order. As a by product we now allow a page to be copied multiple times (By repeating the page index multiple time in the selection)

Illustration of the problem (Copied)

This is my main document
- Root (Page Tree)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 1)
      - Resources
        - Image (F3) (User Icon)
    - Page (Page No. 2)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Another Logo)
      - Font (F2) ( Another Custom Font)
    - Page (Page No. 3)
      - Resources
        - Image (F3) (User Icon)

Now, let's try to merge the selection of pages (1, 3, 2) in that order:

- Root (Page Tree)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 1)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Another Logo)
      - Font (F2) ( Another Custom Font)
    - Page (Page No. 3)
      - Resources
        - Image (F3) (User Icon)
  - Tree (Child Page Tree)
    - Resources
      - Image (F1) (Logo)
      - Font (F2) (Custom Font)
    - Page (Page No. 2)
      - Resources
        - Image (F3) (User Icon)

I solve that by copying the page from bottom to top, instead of top to bottom like it was before. By doing so, we got rid of one recursive method (`DocumentMerger::CopyPagesTree()`). I think thanks to this last bit, we can introduce `Parallelism`, but that's a topic for another day.

**I would like to comment, that I don't like the current state of public PdfMerger API**

**cc** @Poltuu I think this would entirely resolve #244, any comments?

### DIFF
--- a/src/UglyToad.PdfPig/Writer/PdfMerger.cs
+++ b/src/UglyToad.PdfPig/Writer/PdfMerger.cs
@@ -16,6 +16,8 @@
     using Tokenization.Scanner;
     using Tokens;
     using Exceptions;
+    using System.Linq;
+    using Util;
 
     /// <summary>
     /// Merges PDF documents into each other.
@@ -29,7 +31,7 @@
         /// <summary>
         /// Merge two PDF documents together with the pages from <paramref name="file1"/> followed by <paramref name="file2"/>.
         /// </summary>
-        public static byte[] Merge(string file1, string file2)
+        public static byte[] Merge(string file1, string file2, IReadOnlyCollection<int> file1Selection = null, IReadOnlyCollection<int> file2Selection = null)
         {
             if (file1 == null)
             {
@@ -45,7 +47,7 @@
             {
                 File.ReadAllBytes(file1),
                 File.ReadAllBytes(file2)
-            });
+            }, new [] { file1Selection, file2Selection });
         }
 
         /// <summary>
@@ -67,13 +69,13 @@
                 bytes.Add(File.ReadAllBytes(filePath));
             }
 
-            return Merge(bytes);
+            return Merge(bytes, null);
         }
 
         /// <summary>
         /// Merge the set of PDF documents.
         /// </summary>
-        public static byte[] Merge(IReadOnlyList<byte[]> files)
+        public static byte[] Merge(IReadOnlyList<byte[]> files, IReadOnlyList<IReadOnlyCollection<int>> pagesBundle = null)
         {
             if (files == null)
             {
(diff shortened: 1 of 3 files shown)


src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -67,13 +69,13 @@ public static byte[] Merge(params string[] filePaths)
...
 
-            return Merge(bytes);
+            return Merge(bytes, null);
         }
 
         /// <summary>
         /// Merge the set of PDF documents.
         /// </summary>
-        public static byte[] Merge(IReadOnlyList<byte[]> files)
+        public static byte[] Merge(IReadOnlyList<byte[]> files, IReadOnlyList<IReadOnlyCollection<int>> pagesBundle = null)
```
Poltuu, line 78: worth considering, a `IReadOnlyList<(byte[] File, IReadOnlyCollection<int> Pages)>` signature.

Poltuu (commented):
As far as i'm concerned, this looks great.
I can't pretend I know too much on the subject, so someone else should be looking into this

plaisted:
This looks really useful. PdfPig looks nice for text extraction but is missing some features such as splitting / merging documents, modify existing documents, etc. Looks like you've put a lot of work into these areas @InusualZ thanks for the effort and hope these get merged soon.

Poltuu (approved):

src/UglyToad.PdfPig/Writer/PdfMerger.cs
EliotJones, line 78: `IReadOnlyCollection` or `ICollection` are probably the wrong types here, since they do not guarantee order iirc, I'd recommend `IReadOnlyList` since order is important here.

EliotJones (approved):
LGTM, just the doubt on collection vs list types.

PR #249, opened 2020-12-20
Fix TokenWriter::WriteString not escaping properly special characters

InusualZ:
Before this fix we were producing a broken binary:

Evidence

The left document is the original. The right one, is the same document, but the operations of the content stream of that page were written by the library. As you can see operations are being treated as text of the page, rather than operation itself.

ContentStream

Left content stream is from the original document, right content stream come from the document created by the library. As you can see in the we are not properly escaping the special characters. 

The pertinent paragraph in the specifications is:
PDF Spec

I didn't find where you were putting the TokenWriter test so I created the file (I didn't search very much either 😢).
Please comment on anything else...

PD: by any chance we can get right of those extra white spaces?


EliotJones (approved):
LGTM, thanks for the fix

PR #250, opened 2020-12-21
[New Feature] PdfPageBuilder: Allow to copy pages from another document

InusualZ:
This is a very naive implementation, because if you copy a lot of pages from a given document you would have duplicated resources. But, I would like to get some thoughts on this implementation.

I was surprise by the amount of code that this feature needed to be implemented. I was expecting a lot more. Even so, there is some duplicated code in two parts:
- [Copying Token](https://github.com/InusualZ/PdfPig/commit/7126564eefbc30c205a165739d333dd6da3ddb0e#diff-a4cfeaa02b17ecfae6639ce457a744a267bf3e7d478740f850a02d0e42cb8c0eR431-R492) This piece of code keep reappearing, I really don't like it.
- [Resource Management](https://github.com/InusualZ/PdfPig/commit/7126564eefbc30c205a165739d333dd6da3ddb0e#diff-ffb8dc0dfd0b8bc578f2ecd4517965154295f4b8900e71ed5948bc2325dd12eeR537-R666) I think this piece of code and the one that write the resources to the stream can be transformed into a nice interface behind a read only interface and mutable interface, kind of like `IReadOnlyList<>` and `IList<>`, but `IReadOnlyResourceStore` and `IResourceStore`. I would like yours thought about that

Anyway, This code have some `TODO` I would like to address them, but I don't know If you even want the feature in it's current path of implementation, so maybe the effort to implement them are not worth.

If you have any suggestions, thought, anything feel free to comment.

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Writer/PdfDocumentBuilderTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Writer/PdfDocumentBuilderTests.cs
@@ -644,6 +644,53 @@
             }
         }
 
+        [Fact]
+        public void CanCopyPage()
+        {
+
+            byte[] b;
+            {
+                var builder = new PdfDocumentBuilder();
+
+                var page1 = builder.AddPage(PageSize.A4);
+
+                var file = TrueTypeTestHelper.GetFileBytes("Andada-Regular.ttf");
+
+                var font = builder.AddTrueTypeFont(file);
+
+                page1.AddText("Hello", 12, new PdfPoint(30, 50), font);
+
+                Assert.NotEmpty(page1.CurrentStream.Operations);
+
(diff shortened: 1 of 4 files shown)


EliotJones (approved):
This looks like a useful feature to have, I'm just uncertain what the behaviour looks like for the copying a page's resource dictionary where the dictionary is using inherited properties, e.g. fonts, rotation or mediabox are set by the page trees pages node rather than at the page level. I think we can find out later and just start with this approach for now :+1:

For the questions you raise in the opening PR comment I think we're fine taking the current approach for now, I'd avoid trying to derive the abstractions prematurely so I'm fine with doing things in a maybe less efficient manner to begin with.

PR #252, opened 2021-01-06
fix stackoverflow on pdfmerger

Poltuu:
Example found on 68-1990-01_A.pdf:

Obj Ref1
 Kids [Ref2]

Obj Ref2
 Parent Ref1

Let me know if ther is a better/preferred way to handle reference cycles.

### DIFF
--- a/src/UglyToad.PdfPig/Writer/PdfMerger.cs
+++ b/src/UglyToad.PdfPig/Writer/PdfMerger.cs
@@ -415,15 +415,15 @@
                                 return newReferenceToken;
                             }
 
-                            var tokenObject = DirectObjectFinder.Get<IToken>(referenceToken.Data, tokenScanner);
+                            //we add the token to referencesFromDocument to prevent stackoverflow on references cycles 
+                            newReferenceToken = context.ReserveNumberToken();
+                            referencesFromDocument.Add(referenceToken, newReferenceToken);
 
+                            var tokenObject = DirectObjectFinder.Get<IToken>(referenceToken.Data, tokenScanner);
                             Debug.Assert(!(tokenObject is IndirectReferenceToken));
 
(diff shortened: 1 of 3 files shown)


src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -415,15 +415,15 @@ private IToken CopyToken(IToken tokenToCopy, IPdfTokenScanner tokenScanner, IDic
                                 return newReferenceToken;
                             }
 
-                            var tokenObject = DirectObjectFinder.Get<IToken>(referenceToken.Data, tokenScanner);
+                            //we add the token to referencesFromDocument to prevent stackoverflow on references cycles 
+                            newReferenceToken = context.WriteToken(null);
```
InusualZ, line 419: You can reserve a reference by doing `PdfStreamWriter::ReserveNumberToken`, by doing that you don't have to write a null token. I was expecting that `PdfStreamWriter::WriteToken` should not let you write null token, because this would be incompatible with an immediate flusher pdf stream writer
```suggestion
                            //we add the token to referencesFromDocument to prevent stackoverflow on references cycles 
                            newReferenceToken = context.ReserveNumberToken();
                            referencesFromDocument.Add(referenceToken, newReferenceToken);

                            var tokenObject = DirectObjectFinder.Get<IToken>(referenceToken.Data, tokenScanner);
                            Debug.Assert(!(tokenObject is IndirectReferenceToken));

                            var newToken = CopyToken(tokenObject, tokenScanner, referencesFromDocument);
                            return context.WriteToken(newToken, (int)newReferenceToken.Data.ObjectNumber);
```

Poltuu: thanks. i updated the branch

InusualZ (approved):
LGTM

PR #253, opened 2021-01-06
Opening a PdfDocument via a file path opens a stream instead of allocating

Poltuu:
In this PR, I suggest changing the default way to open a file, by switching to streams instead of getting the byte[].

This should be mostly beneficial for large documents, and treatments in loops (less pressure on the GC mostly).

### DIFF
--- a/src/UglyToad.PdfPig/Parser/PdfDocumentFactory.cs
+++ b/src/UglyToad.PdfPig/Parser/PdfDocumentFactory.cs
@@ -40,7 +40,8 @@
                 throw new InvalidOperationException("No file exists at: " + filename);
             }
 
-            return Open(File.ReadAllBytes(filename), options);
+            var input = new StreamInputBytes(File.OpenRead(filename), true);
+            return Open(input, options);
         }
 
         internal static PdfDocument Open(Stream stream, ParsingOptions options)


src/UglyToad.PdfPig/Parser/PdfDocumentFactory.cs
```diff
@@ -40,7 +40,8 @@ public static PdfDocument Open(string filename, ParsingOptions options = null)
                 throw new InvalidOperationException("No file exists at: " + filename);
             }
 
-            return Open(File.ReadAllBytes(filename), options);
+            var input = new StreamInputBytes(File.OpenRead(filename), true);
+            return Open(input, options);
```
EliotJones, line 44: Thanks for your PR(s, I'll try to get to the rest soon).

My only concern here is that streams were less performant than byte arrays in my performance testing due to the need to seek randomly and rewind bytes in the file-stream. PDFBox gets around this by using a buffered stream type.  I wonder if we're better off keeping `byte[]` as default (I'd assume the size distribution of PDF files has a long tail but is generally on the order of kilobytes and having advanced users opt in to slower-but-lower-memory-footprint parsing using the stream open method explicitly. Another possibility is using `BufferedStream` for the frameworks where it is available https://docs.microsoft.com/en-us/dotnet/api/system.io.bufferedstream?view=net-5.0

Let me know your thoughts :)

schmitch: actually we deal a lot with pdf's and we've seen ranges from 0.5mb till 200mb, especially when dealing with scanned pdf's the size distribution can be really really big, because they are only images and don't compress too well.
I think it might be a good idea to look for something like a buffered stream. btw. buffered stream is also available in .net standard 2.0 so it has a wide support range.

EliotJones: I think this should be possible for the consumer to provide themselves, for instance:

```
using var fileStream = File.OpenRead(@"C:\temp\mypdf.pdf");
using var myBuffered = new BufferedStream(fileStream);

using var pdf = PdfDocument.Open(myBuffered);
```

I'm going to close this PR for now, I think the people who are processing full page scans and technical drawings are in a minority that have the larger documents and the API has sufficient affordances for them to use the stream input type where required. I'd need a lot more profiling to be comfortable making this change and I'm not realistically going to have time or resources to do it.

@Poltuu thank you for the suggested change and sorry it didn't go anywhere.

EliotJones:
As mentioned in https://github.com/UglyToad/PdfPig/pull/253#discussion_r687294140 we'll keep the default input types for now unless there's an outcry to revisit.

PR #254, opened 2021-01-06
Pdfmerger stream apis

Poltuu:
This pr does two things:
 - pdfmerger uses stream instead of byte[]. This should improve performance for large documents/treatments in loop.
 - adds an optional output Stream argument on PdfMerger methods. This prevents allocating the newly created documents in memory.


EliotJones (approved):
Looks good to me, thanks for the PR

Poltuu:
Conflicts fixed

PR #257, opened 2021-01-07
Allow to have empty code space, fixes #256

### LINKED ISSUE #256: InvalidOperationException : Codespace range contains an unexpected token
Hi,

Would you tell me why an exception is thrown on the GetWords() of this page [pdfpig_error_CodespaceRangeParser_page.pdf](https://github.com/UglyToad/PdfPig/files/5782438/pdfpig_error_CodespaceRangeParser_page.pdf)

Have a good day.

InusualZ:
Like the commit message say, allow to _parse_ empty code spaces, that way font that have `ToUnicode` dictionary, but no code spaces can be parsed correctly.

PDF Specification related to this problem. I'm in no expert in font, so this may not be the correct fix.
PDF Specs


EliotJones (approved):
Thanks for handling this while I'm (more-or-less) away, looks good to me.

PR #259, opened 2021-01-14
PdfMerger: Fix indirect object lookup dict

plaisted:
Current implementation uses `IndirectReferenceToken` as the dictionary key which does not override GetHashCode or Equals resulting TryGetValue failing even though matching indirect objects already have been inserted. Modified to use the `IndirectReference` as the key instead and added test coverage to verify indirect objects are not duplicated from same document.


EliotJones (approved):
Thanks, this looks like a good fix

PR #262, opened 2021-01-15
Pdfmerger complete rearrangment api

Poltuu:
In this PR, I introduce two new public apis:

 - `PdfMerger.Merge(IReadOnlyList<Stream> streams, Stream output, IPdfArrangement pageArrangment)`: this uses the new `IPdfArrangement` that lets the consumer entirely describe the way the new pdf is supposed to be generated, allowing new things like zip-merge for instance.

 - `PdfMerger.MergeMany(IReadOnlyList<Stream> streams, IReadOnlyCollection<(Stream Output, IPdfArrangement PageArrangments)> exports)` : similar api but for multiple output. Useful for pdf split.

I followed comments from [this PR](https://github.com/UglyToad/PdfPig/pull/251) and did not add other specialized apis (PdfSplitter.Split, PdfEditor.RemovePages), but I feel like it would be a good idea. I believe those two new apis would benefit from being exposed on a different class, something like `PdfCombiner` ou `PdfArranger`.

I also did not expose different `IPdfArrangement` implementation that would simplify differents specialized kind of merge, but I believe that could also be a way to simplify the api.

### DIFF
--- a/src/UglyToad.PdfPig.Tests/PublicApiScannerTests.cs
+++ b/src/UglyToad.PdfPig.Tests/PublicApiScannerTests.cs
@@ -211,6 +211,7 @@
                 "UglyToad.PdfPig.Writer.PdfMerger",
                 "UglyToad.PdfPig.Writer.PdfPageBuilder",
                 "UglyToad.PdfPig.Writer.TokenWriter",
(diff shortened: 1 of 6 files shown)


Poltuu:
replaced by this https://github.com/UglyToad/PdfPig/pull/279

PR #263, opened 2021-01-15
Use Euclidean distance in PdfRectangle's width and height

### LINKED ISSUE #261: Why is the Height of PdfRectangle is negative when Top is greater than Bottom?
PdfRectangle rectangle = new PdfRectangle(left, bottom, top, right);
where 
		left	401.51	double
		bottom	291.45	double
		top	300.17322363281249	double
		right	461.803424	double
result to Height: -170.353424
rectangle

I see the definition of PdfRectangle in Version 0.1.4.0 as follows:
        //
        // Summary:
        //     Create a new UglyToad.PdfPig.Core.PdfRectangle.
        //
        // Parameters:
        //   x1:
        //     Bottom left point's x coordinate of the rectangle.
        //
        //   y1:
        //     Bottom left point's y coordinate of the rectangle.
        //
        //   x2:
        //     Top right point's x coordinate of the rectangle.
        //
        //   y2:
        //     Top right point's y coordinate of the rectangle.
        public PdfRectangle(double x1, double y1, double x2, double y2);

Please help to clarify my doubt, thanks in advance!

BobLd:
fix #261

The problem was that after inverse rotation of the points, some where "outside" the page, leading to negative coordinate.

We now use 1D euclidian distance (Math.Abs())

### DIFF
--- a/src/UglyToad.PdfPig.Core/PdfRectangle.cs
+++ b/src/UglyToad.PdfPig.Core/PdfRectangle.cs
@@ -49,6 +49,7 @@
         private double width;
         /// <summary>
         /// Width of the rectangle.
+        /// <para>A positive number.</para>
         /// </summary>
         public double Width
         {
@@ -66,6 +67,7 @@
         private double height;
         /// <summary>
         /// Height of the rectangle.
+        /// <para>A positive number.</para>
         /// </summary>
         public double Height
         {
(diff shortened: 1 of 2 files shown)


EliotJones (approved):
LGTM, thanks for fixing

BobLd:
Merged

PR #269, opened 2021-01-20
Remove locks / SB pooling during CoreTokenScanner parsing

plaisted:
This seems to be the simplest way remove the locking contention when multiple instances of PdfDocument are used in the same program in parallel.

The tokenizers are never used in parallel for a given `CoreTokenScanner`. Converting them to non-static and having each Tokenizer contain their own StringBuilder removes the need for the `StringBuilderPool` and performance is improved significantly in parallelized parsing scenarios (~40% improvement in my basic testing using a DOP of 4, would only grow as DOP was higher). In non-parallel scenarios performance difference should improve slightly as it has no locking for every string/number (and only allocates a few extra objects per page).


EliotJones (approved):
Thanks, this should definitely improve performance in some common use-cases

PR #270, opened 2021-01-21
[New Features] Replace token, trim non-text streamcontent.

plaisted:
This PR adds two new capabilities:
* Adds the ability to replace the value of `IndirectReference` in a `PdfDocument` with new `IToken`. This is implemented by simply adding an in memory dictionary to the `PdfTokenScanner` for replaced tokens. Currently these tokens are only read when data is subsequently requested from the `PdfTokenScanner` however further work could allow the `PdfDocument` to be saved with the new token. This would allow scenarios such as tweaking existing documents (replace an existing font with another one, embedding it for example). This feature is exposed through the `AdvancedDocumentAccess` property of the `PdfDocument`.
* Adds the ability to trim all content streams in a PDF document to remove all non-text operations. This feature is exposed through the `AdvancedDocumentAccess.StripNonText` method. It uses the first feature to replace the content stream objects in the document with new versions that only contain text operations. The rest of the library is unaffected. If the method is called prior to requesting a page using `PdfDocument.GetPage()` the parser will read the replaced token and only have to parse the text contents. This can dramatically increase text extraction performance of PDFs with large amounts of graphic operations (eg. vector graphics, tables, lines, etc). In large graphic heavy PDFs I've observed extraction durations reduced by 50-75%. In "normal" PDFs I've generally seen 15-30% performance increase. In an text only PDF this feature will increase total parsing time slightly as it pre-processes the content stream during text stripping (removing nothing in a text only PDF). As the decrease in performance in these cases is very small and the feature is opt in (must call AdvancedDocumentAccess.StripNonText to activate) this shouldn't be a concern.


InusualZ:
I have mix feeling about this feature. Have you tried to instead of removing/replace operations, just ignore them while the content stream is being parsed?

plaisted:
It's not really possible without major changes to the existing setup since the current parsing process allocates objects for everything as it's parsing. Even ignoring them it would still take a pretty big performance hit and then we'd have to add logic into parsing to ignore other operations.  The way it's implemented right now, they entire parsing flow is unaffected.

The important feature is really being able to modify existing indirect objects of the PDF. The rest of it could be completed in an external library I just thought it was useful enough to include as the library is very text focused and I'm seeing ~50% performance improvements (after my latest tweaks) for most PDFs in my test suite using the text only mode.

Issue #258, opened 2021-01-12
IndexOutOfRangeException - data.Seek(position) - NameTableParser.cs


yvainb:
Hello

I get the error `System.IndexOutOfRangeException: Index was outside the bounds of the array` when calling the function `document.GetPage(i).GetWords():`

It is caused in the `NameTableParser.cs` file line 75 (data.Seek(position);)
More precisely in the file `ByteArrayInputBytes.cs` line 97:
![pdfpig_error_ByteArrayInputBytes cs_line97](https://user-images.githubusercontent.com/22787451/104301723-4273a680-54c8-11eb-93ec-588d02701434.png)

The error can be reproduced with this page : [pdfpig_error_data.Seek(position)_page.pdf](https://github.com/UglyToad/PdfPig/files/5801304/pdfpig_error_data.Seek.position._page.pdf)

Would you please explain the problem to me? 

Kind regards,

InusualZ:
Hello, thanks for bringing this issue to our attention, I would take a look at it later today.

Poltuu:
I've looked into it and it seems to me that the "name" table of the embedded fontfile is simply unreadable (starts good, then starts failing at the 7th name). The checksum is correct thus.
As this is not necessary for most application, I'm guessing a solution would be to try catch the NameTableParser.Parse method, but that's only correct if we agree that the file is indeed broken.

I open a PR reflecting my reflection on the matter here https://github.com/UglyToad/PdfPig/pull/272

PR #272, opened 2021-01-25
Fix buggy font

Poltuu:
Fixes https://github.com/UglyToad/PdfPig/issues/258 by try/catch. Only valid if confirmed that the input can be buggy, and that the code os not faulty (which is the case as far as I could tell)

I also add a checksum check, which should be added I guess.


EliotJones:
Hi there, thanks for the fix, I tried to update this branch to fix a couple of the test failures but couldn't work out how to get Git to use my right, non-work credentials to push the changes, I'm going to close this one and merge it under #283

PR #279, opened 2021-02-06
Add pages from PdfDocument to PdfDocumentBuilder

plaisted:
- Adds performant way to copy pages from `PdfDocument` to `PdfDocumentBuilder`, should be similar to PdfMerger performance.
- Standardizes internal PDF writing to use `IPdfStreamWriter`.
- Adds `PdfDedupStreamWriter` which dedups indirect objects as it is writing. This works across different documents. Can be accessed using `PdfDocumentBuilder` constructor.
- Adds `IPageContentStream` abstraction to allow `PdfPageBuilder` to work with raw copied streams that aren't parsed.
- Adds ability to replace tokens in an existing `PdfDocument` before parsing pages or copying them to a `PdfDocumentBuilder`
- Move CopyToken logic to WriterUtil and improve logic to only reserve object numbers on circular references.
- Made some changes to the way `PdfDocumentBuilder` and `PdfPageBuilder` handle fonts to scope names per page instead of globally. Allows less collisions from imported resources.
- Changed `PdfDocumentBuilder` to output pages in a tree, testing showed Acrobat reader performed very poorly with large documents that had a "flat" document layout as the `PdfDocumentBuilder` originally created. Currently hardcoded to use leaf size of 25 pages. Acrobat now very responsive with large documents produced by `PdfDocumentBuilder` (20,000+ pages).
- Switched `PdfMerger` to use `PdfDocumentBuilder` for merging.


plaisted:
@EliotJones From my point of view this is ready.  It cleans up / unifies the document modification approaches to all use `PdfDocumentBuilder` and improves some of the capabilities of `PdfDocumentBuilder`. Let me know what you think.

src/UglyToad.PdfPig/Writer/PdfDocumentBuilder.cs
```diff
@@ -152,21 +193,15 @@ public AddedFont AddStandard14Font(Standard14Font type)
...
+        internal IndirectReferenceToken AddImage(DictionaryToken dictionary, byte[] bytes)
         {
-            var reserved = context.ReserveNumber();
-
-            var stored = new ImageStored(dictionary, bytes, reserved);
-
-            images[stored.Id] = stored;
-
-            return new IndirectReference(reserved, 0);
+            var streamToken = new StreamToken(dictionary, bytes);
```
Poltuu, line 203: is `images` useless now ? or should it still hold the stored images

src/UglyToad.PdfPig/Writer/PdfDocumentBuilder.cs
```diff
@@ -235,275 +270,392 @@ public PdfPageBuilder AddPage(PageSize size, bool isPortrait = true)
...
+        private void CompleteDocument()
+        {
+            // write fonts to reserved object numbers
+            foreach (var font in fonts)
+            {
+                font.Value.FontProgram.WriteFont(context, font.Value.FontKey.Reference);
+            }
 
-                var pagesDictionaryData = new Dictionary<NameToken, IToken>
+            var procSet = new List<NameToken>
```
Poltuu, line 457: this could be static somewhere

src/UglyToad.PdfPig/Writer/PdfPageBuilder.cs
```diff
@@ -509,13 +534,12 @@ public AddedImage AddPng(Stream pngStream, PdfRectangle placementRectangle)
         /// <param name="srcPage">Page to be copied</param>
         public void CopyFrom(Page srcPage)
```
Poltuu, line 535: I guess it is too late, but it feels like this api should use the new logic. I'm afraid one might end up outdated pretty quickly

src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -122,369 +110,39 @@ public static void Merge(IReadOnlyList<Stream> streams, Stream output, IReadOnly
...
                     {
-                        var dictionary = node.NodeDictionary;
-                        if (dictionary.TryGet(NameToken.Resources, tokenScanner, out DictionaryToken resourcesDictionary))
+                        foreach (var i in pages)
                         {
-                            foreach (var pair in resourcesDictionary.Data)
-                            {
-                                resources.Add(pair.Key, CopyToken(pair.Value, tokenScanner, referencesFromDocument));
-                            }
+                            document.AddPage(existing, 1);
```
Poltuu, line 140: this should be i

src/UglyToad.PdfPig/Writer/PdfMerger.cs
```diff
@@ -122,369 +110,39 @@ public static void Merge(IReadOnlyList<Stream> streams, Stream output, IReadOnly
...
-                        if (dictionary.TryGet(NameToken.Resources, tokenScanner, out DictionaryToken resourcesDictionary))
+                        for (var i = 1; i <= existing.NumberOfPages; i++)
                         {
-                            var nonCollidingResources = resourcesDictionary.Data.Keys.Except(resources.Keys);
-                            if (nonCollidingResources.Count() != resourcesDictionary.Data.Count)
-                            {
-                                // This means that at least one of the resources collided
-                                return true;
-                            }
+                            document.AddPage(existing, 1);
```
Poltuu, line 134: this should be i

Poltuu (commented):
Brilliant, really nice that the pdfmerger got to use the PdfDocumentBuilder

src/UglyToad.PdfPig/Writer/PdfDocumentBuilder.cs
```diff
@@ -63,6 +64,47 @@ public class PdfDocumentBuilder
         /// </summary>
         internal IReadOnlyDictionary<Guid, ImageStored> Images => images;
 
+        
+        /// <summary>
+        /// Creates a document builder keeping resources in memory.
+        /// </summary>
+        public PdfDocumentBuilder()
+        {
+            context = new PdfStreamWriter(new MemoryStream(), true);
```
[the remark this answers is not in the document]
plaisted: I think this should be already possible through the `public PdfDocumentBuilder(Stream stream, bool disposeStream=false, PdfWriterType type=PdfWriterType.Default, decimal version=1.7m)` constructor, maybe a simpler one with just the stream would be useful.

src/UglyToad.PdfPig/Writer/PdfPageBuilder.cs
plaisted: I agree, I didn't want to change much there as it was just added by @InusualZ . It can serve a different purpose (allows pages to be overlayed on each other, allows modification of graphics operations after import) but would be nice to reduce some of the duplication.

src/UglyToad.PdfPig/Writer/PdfMerger.cs
plaisted: good catch, adding test case for merger that will cover.

src/UglyToad.PdfPig/Writer/PdfDedupStreamWriter.cs
```diff
@@ -0,0 +1,200 @@
...
+    internal class PdfDedupStreamWriter : IPdfStreamWriter
+    {
+
+        public Stream Stream { get; }
+        private int CurrentNumber { get; set; } = 1;
+        private bool DisposeStream { get; set; }
+        private const decimal DefaultVersion = 1.2m;
+        private bool Initialized { get; set; }
+        private readonly Dictionary<IndirectReference, long> offsets = new Dictionary<IndirectReference, long>();
+        private readonly Dictionary<byte[], IndirectReferenceToken> hashes = new Dictionary<byte[], IndirectReferenceToken>(new FNVByteComparison());
```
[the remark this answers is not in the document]
plaisted: The deduping works by comparing the byte[] data of indirect objects excluding the startobj / endobj header and footer. It does this across documents so `IToken` would be tricky as it would need to be serialized each time.

The FnvHash as key also wouldn't work as it's just a hashcode (integer) and collisions are expected. Internally the dictionary is using the FnvHash code to do the lookups (custom IEqualityComparer).  Eventually we could use a real hash of the byte[] as key but would need to determine how to handle collisions, with this setup collisions aren't possible as it does a full byte[] comparison of the data.

src/UglyToad.PdfPig/Writer/PdfDocumentBuilder.cs
plaisted: Yea, they aren't needed on the PdfDocumentBuilder anymore, will clean this up.

InusualZ (changes requested):
Hello, I have seen that you have been busy :)

I like this PR, specially the part about the `PdfMerger` using the `PdfDocumentBuilder`. This is going to be my first part review since you changed quite a lot of thing. But, from what I have seen this looks pretty good. I would like to take a closer look at how you are handling resources, plus I would like to see how this implementation path could be further improve in later PR.

Anyway, thanks

EliotJones (approved):
Thanks for this work, let's get this merged and then test it a lot and continue to enhance the builder functionality.

Would you mind quickly summarising any public API breaking changes so that we can provide migration notes for the next release version please?

plaisted:
@EliotJones I did a scan over the changes between this PR and #250 and it looks like the only two breaking changes are:

- `PdfDocumentBuilder` now implements `IDisposable`. This disposes the underlying stream by default but this is a MemoryStream normally so not any serious consequences if left undisposed.
- PdfPageBuilder had the `AdvancedEditing` property removed. The same functionality seems to be available in the `ContentStream` methods / properties (this was from #250)

PR #282, opened 2021-02-07
Partial optional content (aka layers collections) extraction

BobLd:
Partial optional content extraction: Add `page.GetOptionalContents()` and add tests.

More work needs to be done to take in account more cases, like:
-  ability to process XObjects and annotations
- ability to process OCMD type

More info in Pdf Ref 1.7, section 4.10


BobLd:
Marked contents need to be processed recursively, in order to check if children also have the OC tag

src/UglyToad.PdfPig/Content/Page.cs
```diff
@@ -193,6 +194,42 @@ public IEnumerable<Annotation> GetAnnotations()
             {
                 return annotationProvider.GetAnnotations();
             }
+
+            /// <summary>
+            /// Gets any optional content on the page.
+            /// <para>Does not handle XObjects and annotations for the time being.</para>
+            /// </summary>
+            public IDictionary<string, IReadOnlyList<OptionalContentGroupElement>> GetOptionalContents()
```
EliotJones, line 202: I'd prefer `IReadOnlyDictionary<string, IReadOnlyList<OptionalContentGroupElement>>` since it is explicit that mutating the return type will not change anything on the `page` object

src/UglyToad.PdfPig/Content/PageContent.cs
```diff
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
```
EliotJones, line 5: Is this using needed?

src/UglyToad.PdfPig/Content/OptionalContentGroupElement.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Underlying <see cref="MarkedContentElement"/>.
+        /// </summary>
+        public MarkedContentElement MarkedContent { get; }
+
+        internal OptionalContentGroupElement(MarkedContentElement markedContentElement)
+        {
+            MarkedContent = markedContentElement;
+
+            // Type - Required
```
EliotJones, line 44: We might need to store an IPdfTokenScanner internally here and use the TryGetv overload that takes in the scanner since these can be indirect references I assume?

EliotJones (approved):
LGTM for now, a few minor things to adjust

src/UglyToad.PdfPig/Content/PageContent.cs
BobLd: will check

src/UglyToad.PdfPig/Content/OptionalContentGroupElement.cs
BobLd: will check that too

PR #284, opened 2021-02-08
fixes NullToken for a font

Poltuu:
A font was in a streamObject, PdfTokenScanner.ParseObjectStream would return a NullToken (as the object was indeed null)

DirectObjectFinder.Get would then throw on line 79, even thus it should handle null tokens (i.e. ResourceStore.LoadFontDictionary line 152).

I'm unfortunately unable to provide the original pdf, for privacy reasons.

Being able to quickly have a release with this fix would be greatly appreciated 🙏


EliotJones (approved):
LGTM

Poltuu:
thanks

Issue #271, opened 2021-01-22
Fail to extract image from pdf


oleksant:
Hi. 

During extraction of images from the pdf files we keep running into failures. 
Up to 30 percent of images fail to extract. 

We have tried TryGetPng, TryGetBytes, and RawBytes for sureб but some of the images could not be extracted anyway. 

To create the PDF file from Word we are using LibreOffice. 

Here is a piece of code that is using RawBytes. It saves the image but the result could not be opened.
 
    using (var document = PdfDocument.Open(FilePath))
            {
                foreach (var page in document.GetPages())
                {
                    foreach (var image in page.GetImages())
                    {
                        var byteArray = image.RawBytes.ToArray();
                        using (var fs = new FileStream(imageFilePath, FileMode.Create, FileAccess.Write))
                        {
                            fs.Write(byteArray, 0, byteArray.Length);
                        }
                    }
                }
            }

Usually, I can check whether the image fails to extract with the following code: 

```
public static bool IsValidImage(byte[] bytes)
{
      try {
            using(MemoryStream ms = new MemoryStream(bytes))
                    Image.FromStream(ms);
            }
            catch (ArgumentException) {
                return false;
            }
            return true; 
}
```

[StuckSamplePdf.pdf](https://github.com/UglyToad/PdfPig/files/5855093/StuckSamplePdf.pdf)
[TextImages.pdf](https://github.com/UglyToad/PdfPig/files/5855596/TextImages.pdf)
[2imagesdoc.pdf](https://github.com/UglyToad/PdfPig/files/5855597/2imagesdoc.pdf)

alex19497:
After some investigation, I can assume that for most images (except for the one in StuckSamplePdf) we are facing the same issue. 
If the ColorSpace is an array - we have an issue. In our case - the name of the first element is "Indexed" and from the PdfBox code, I can see that it requires additional logic, saving other data from the ColorSpace array to be able to decode the image with it.

Here is the data from our ColorSpace: 
![image](https://user-images.githubusercontent.com/44353204/105523327-ed653b00-5ce6-11eb-802c-c8ecca734116.png)

The 3 element is never being used and it is needed to reconstruct the original image.

alex19497:
Hi again. There is no mechanism to process raster images in PdfPig as I can see, so in our case, the workaround that we were able to put in place is the following:

```
 public static class PdbPigImageExtentions
    {
[...]

PR #285, opened 2021-02-08
Support indexed color space

EliotJones:
For #271 add support for retrieving associated ColorSpace information and util method `ColorSpaceDetailsByteConverter.Convert` for converting indexed data.


src/UglyToad.PdfPig/Images/ColorSpaceDetailsByteConverter.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        private static byte[] UnwrapIndexedColorSpaceBytes(IndexedColorSpaceDetails indexed, IReadOnlyList<byte> input)
+        {
+            var multiplier = 1;
+            Func<byte, IEnumerable<byte>> transformer = null;
+            switch (indexed.BaseColorSpaceDetails.Type)
+            {
+                case ColorSpace.DeviceRGB:
+                    transformer = x =>
```
Poltuu, line 49: it seems to me like this sort of code should be on an abstract method on the `ColorSpaceDetails` class :)

EliotJones: Yes, I'm sort of on the fence about it, in general I'm trying to aim for keeping data and behaviour entirely separate, so data classes just encapsulate their data and enforce invariants/immutability while classes like this contain the behaviour as pure functions.

In this case though it probably makes more sense to keep the behaviour with the data, which is more traditional C# style, but I worry about ending up with a traditional C# codebase with that approach 😆

PR #308, opened 2021-03-17
add ccittfax implementation

mind-ra:
Pull request for ccittfax implementation as @EliotJones taught me.

### DIFF
--- a/src/UglyToad.PdfPig/Filters/CcittFaxDecodeFilter.cs
+++ b/src/UglyToad.PdfPig/Filters/CcittFaxDecodeFilter.cs
@@ -4,12 +4,13 @@
     using System.Collections.Generic;
     using System.Linq;
     using Tokens;
+    using UglyToad.PdfPig.Util;
 
     internal class CcittFaxDecodeFilter : IFilter
     {
         /// <inheritdoc />
         public bool IsSupported { get; } = true;
-        
+
         /// <inheritdoc />
         public byte[] Decode(IReadOnlyList<byte> input, DictionaryToken streamDictionary, int filterIndex)
         {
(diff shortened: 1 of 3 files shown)


src/UglyToad.PdfPig/Filters/CcittFaxDecodeFilter.cs
```diff
@@ -2,18 +2,26 @@
...
-        public bool IsSupported { get; } = false;
-
+        public bool IsSupported { get; } = true;
+        
         /// <inheritdoc />
         public byte[] Decode(IReadOnlyList<byte> input, DictionaryToken streamDictionary, int filterIndex)
         {
-            throw new NotSupportedException("The CCITT Fax Filter for image data is not currently supported. " +
-                                            "Try accessing the raw compressed data directly.");
+            if (input == null)
```
EliotJones, line 16: I think the expected output here would be to convert the image data to a bitmap similar to https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/filter/CCITTFaxFilter.java

I'll merge this change but I think I'll remove this filter implementation and update the PdfImageHelper implementation to be (see next comment)

src/UglyToad.PdfPig/Util/PdfImageHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            if (image == null || image.GetType() != typeof(XObjectImage))
+            {
+                return false;
+            }
+            if(!image.TryGetBytes(out var lazyBytes))
+            {
+                lazyBytes = image.RawBytes;
+            }
+            var pureBytes = lazyBytes.ToArray();
```
EliotJones, line 36: Here after removing the CcittFaxFilter implementation I think we want `var pureBytes = image.RawBytes;` since those are just the TIFF bytes anyway, but beforehand we want to check whether the `xImage.ImageDictionary` filters contains the CcittFax filter, I'll write up an implementation of this after I merge this PR.

PR #312, opened 2021-04-02
fix null token handling in document creation

Poltuu:
Prevent a nullreference exception when the font in the dictionary is a `NullToken` (valid case).

I'm afraid I am unable to provide a test for this, as I can't share the pdf.


EliotJones (approved):
Thanks for the fix

PR #316, opened 2021-04-26
CMYK has four color components

kasperdaff:
PngFromPdfImageFactory.TryGenerate erroneously assumes that CMYK images consist of three color components, which makes it determine that the bytesPure array is incorrectly sized.


EliotJones (approved):
Thanks for the fix!

Issue #313, opened 2021-04-07
incorrect text encoding for string tokens


schmitch:
currently string tokens sometimes have an incorrect encoding. sadly I can't share documents, but if I call:

```csharp
Console.WriteLine(document.Information.Author);
```

the string that will be printed sometimes start with `0xFE`, `0xFF`. btw. apache-pdfbox does this correctly.
Also I can print the following:

```csharp
if (document.Information.DocumentInformationDictionary.TryGet(NameToken.Author, out var token))
{
    var sToken = (StringToken)token;
    var bytes = sToken.GetBytes();
    Console.WriteLine($"AuthorToken: |{bytes[0] & 0xFF}, {bytes[1] & 0xFF}|");
}
```

which will result in: `AuthorToken: |254, 255|` in case of apache-pdfbox the `getString()` method of these tokens will be like that:

```java
    public String getString()
    {
        // text string - BOM indicates Unicode
        if (bytes.length >= 2)
        {
            if ((bytes[0] & 0xff) == 0xFE && (bytes[1] & 0xff) == 0xFF)
            {
                // UTF-16BE
                return new String(bytes, 2, bytes.length - 2, Charsets.UTF_16BE);
            }
            else if ((bytes[0] & 0xff) == 0xFF && (bytes[1] & 0xff) == 0xFE)
            {
                // UTF-16LE - not in the PDF spec!
                return new String(bytes, 2, bytes.length - 2, Charsets.UTF_16LE);
            }
        }
        // otherwise use PDFDocEncoding
        return PDFDocEncoding.toString(bytes);
    }
```
this method will also be called when calling `document.getDocumentInformation().getAuthor()` since the `getAuthor()` method calls `info.getString( COSName.AUTHOR );` which calls:

```java
    public String getString(COSName key)
    {
        String retval = null;
        COSBase value = getDictionaryObject(key);
        if (value instanceof COSString)
        {
            retval = ((COSString) value).getString();
        }
        return retval;
    }
```

so the `StringToken.Data` of PdfPig will always incorrectly also spit out the BOM.

I have no idea why this is happening, since `StringTokenizer` looks correct?!

**Edit:**
[...]

PR #321, opened 2021-04-30
correctly detect encrypted string encoding when decrypting #313

EliotJones:
https://github.com/UglyToad/PdfPig/issues/313#issuecomment-829266556

### DIFF
--- a/src/UglyToad.PdfPig/Encryption/EncryptionHandler.cs
+++ b/src/UglyToad.PdfPig/Encryption/EncryptionHandler.cs
@@ -402,7 +402,7 @@
 
                         var decrypted = DecryptData(data, reference);
 
-                        token = new StringToken(OtherEncodings.BytesAsLatin1String(decrypted));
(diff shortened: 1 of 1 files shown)


EliotJones:
> 
> 
> I was thinking on the same approach. Then, I thought that some bytes could be dropped because the [encoding](https://github.com/UglyToad/PdfPig/blob/c74ca5fda8a51f3af9cf19c675272f57d3beee60/src/UglyToad.PdfPig/Encryption/EncryptionHandler.cs#L401) would not recognize them as valid _codepoint_. I was worried that, that could be the case for `0xFFFE` and `0xFEFF`.
> 
> Maybe I'm overthinking the problem...

I *think* the encoding preserves them, as far as I can tell the same approach of oscillating between strings and binary data is used for the non-encrypted string tokens and I think it preserves the BOM without issues.

InusualZ (approved):
LGTM

PR #323, opened 2021-05-05
Improves support for indexed colorspace images.

kasperdaff:
For indexed color space images where BitPerComponents < 8, the indices need to be 'unpacked' from the individual bytes.


EliotJones (approved):
Thanks for the fixes and the good tests

PR #326, opened 2021-05-10
Fixes for recent CCITT Fax Filter implementation

kasperdaff:


EliotJones (approved):
Thanks for fixing the implementation issues

PR #328, opened 2021-05-11
Fixes null reference exception

d-franklin:
Fixes an issue I was having reading a PDF when encodingDictionary is null.

``System.NullReferenceException: Object reference not set to an instance of an object.
         at UglyToad.PdfPig.PdfFonts.Parser.EncodingReader.ReadEncodingDictionary(DictionaryToken encodingDictionary, Encoding fontEncoding) in E:\PdfPig\src\UglyToad.PdfPig\PdfFonts\Parser\EncodingReader.cs:line 66``


EliotJones (approved):
Thanks for the fix

PR #331, opened 2021-05-30
Various image related fixes and ICCBased (limited), CalGray and CalRGB color spaces support

kasperdaff:


src/UglyToad.PdfPig/Util/ColorSpaceDetailsParser.cs
```diff
@@ -98,22 +107,159 @@ internal static class ColorSpaceDetailsParser
...
+                            !dictionaryToken.TryGet(NameToken.WhitePoint, out ArrayToken whitePointToken))
+                        {
+                            return UnsupportedColorSpaceDetails.Instance;
+                        }
+
+                        var whitePoint = whitePointToken.Data.OfType<NumericToken>().Select(x => x.Data).ToList();
+
+                        // BlackPoint is optional
+                        IReadOnlyList<decimal> blackPoint = null;
+                        if (dictionaryToken.TryGet(NameToken.BlackPoint, out ArrayToken blackPointToken))
```
EliotJones, line 181: I think for all calls to `DictionaryToken.TryGet` in this class/method we'd need to use the overload/extension which takes `IPdfTokenScanner` as the second argument since it's quite likely these dictionary values will be behind indirect references (e.g. `103 0 R` where object 103 is an array), so for example this call would become:

```suggestion
                        if (dictionaryToken.TryGet(NameToken.BlackPoint, scanner, out ArrayToken blackPointToken))
```

This way it should automatically look up the actual ArrayToken even where it is not directly embedded in the dictionary, which seems to happen frequently. I often forget it myself and then get bug reports where the dictionary contains an indirect reference.

EliotJones (commented):
Thanks, you've done so much work here, looks great. Just a minor comment about retrieving dictionary values then we should be good to go.

kasperdaff:
Done

EliotJones (approved):

kasperdaff:
@EliotJones Forgot to merge this PR? 😄

PR #335, opened 2021-06-13
PdfTokenScanner: Fix reading stream operator

InusualZ:
While testing the library wth a bunch of pdf. I found this "error". The pdf is readable by Adobe Reader and Foxit Reader, although the pdf is not completely compliant since after the `stream` operator contain a space.

Stream Operator

So I make the scanner more tolerable to `garbage` after the operator.

The second commit is not related to this problem at all, but it would allow us to see more clearly what token we are reading vs what token we were expecting. QoL

### DIFF
--- a/src/UglyToad.PdfPig/Tokenization/Scanner/PdfTokenScanner.cs
+++ b/src/UglyToad.PdfPig/Tokenization/Scanner/PdfTokenScanner.cs
@@ -256,35 +256,20 @@
 
             // Verify again that we start with "stream"
             var hasStartStreamToken = ReadStreamTokenStart(inputBytes, startStreamTokenOffset);
-
             if (!hasStartStreamToken)
             {
                 return false;
             }
 
             // From the specification: The stream operator should be followed by \r\n or \n, not just \r.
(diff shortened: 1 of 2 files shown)


src/UglyToad.PdfPig/Tokenization/Scanner/PdfTokenScanner.cs
```diff
@@ -256,35 +256,23 @@ private bool TryReadStream(long startStreamTokenOffset, bool getLength, out Stre
...
-            }
-
-            if (inputBytes.CurrentByte != '\n')
-            {
-                if (!hadWhiteSpace)
+                if (!inputBytes.MoveNext())
                 {
                     return false;
                 }
+            } while (char.IsWhiteSpace((char)inputBytes.CurrentByte));
```
EliotJones, line 272: I think we specifically want to stop on the first `\n` here to be compliant with the spec since the data could start with any whitespace, for example where 0x20 is whitespace:

```
stream0x200x20\r\n
0x205ay36
```

EliotJones (approved):
Sorry for the delay

PR #338, opened 2021-06-21
Support for JBIG2 (port of https://github.com/apache/pdfbox-jbig2)

kasperdaff:


EliotJones:
Thanks for this contribution. I'm probably not going to get time to do a review until next weekend, sorry for the delay.

src/UglyToad.PdfPig/Filters/Jbig2/ValueNode.cs
```diff
@@ -0,0 +1,45 @@
...
+                return (rangeLow - iis.ReadBits(rangeLength));
+            }
+            else
+            {
+                // B.4 5)
+                return rangeLow + iis.ReadBits(rangeLength);
+            }
+        }
+
+        internal static string bitPattern(int v, int len)
```
EliotJones, line 36: Java naming 🙈

src/UglyToad.PdfPig/XObjects/XObjectFactory.cs
```diff
@@ -30,15 +29,15 @@ internal static class XObjectFactory
...
             }
 
-            var dictionary = xObject.Stream.StreamDictionary;
+            var dictionary = xObject.Stream.StreamDictionary.Resolve(pdfScanner);
 
             var bounds = xObject.AppliedTransformation.Transform(new PdfRectangle(new PdfPoint(0, 0), new PdfPoint(1, 1)));
 
-            var width = dictionary.Get<NumericToken>(NameToken.Width, pdfScanner).Int;
-            var height = dictionary.Get<NumericToken>(NameToken.Height, pdfScanner).Int;
+            var width = dictionary.Get<NumericToken>(NameToken.Width).Int;
```
EliotJones, line 36: What's the reason for removing the overload using `pdfScanner` in this file? In general overloads of `Get` and `TryGet` parsing `pdfScanner` or `DirectObjectFinder.Get` are required because most tokens in PDF documents can be indirect references to other tokens, in this case an Indirect reference to a number token in its own object.

Usually where they are used it's because there was already a bug in the code which meant they needed to be added, I just want to check we're not causing a regression in this file 🙏

EliotJones (commented):
Thanks for all the hard work here and sorry it has taken so long to review, it's a bit large for one person to feasibly review but other than scanning for obvious security flaws I'm not too worried about any style quibbles since it's all pseudo-Java-ish but isolated in the JBIG folders.

The only comments I'd prefer to address prior to merging are the 2 test files with the scary licensing terms and the changes in XObjectFactory.

BobLd:
I can implement the necessary changes if need be, also saw that it was using `System.Drawing` for the `Rectangle` class, can easily substitute that by `PdfRectangle`

BobLd:
@EliotJones @kasperdaff as it would be sad to lose this work, I will create a new PR by cherry picking the work and adding the changes

BobLd:
Closing PR as it was made available through https://github.com/BobLd/UglyToad.PdfPig.Filters.Jbig2.PdfboxJbig2

Issue #344, opened 2021-07-02
Adobe Reader cannot open some pdf created with PdfPig


fviollette:
Hi, 

I'm using pdfPig to cut out one big Pdf file into several pdf. 
The code below is an example of how I am using your library to create one new pdf with only the first page of my source file : 

```
            var stream = File.OpenRead(pdfPath);
            var sourceDocument = PdfDocument.Open(stream);
            var documentBuilder = new PdfDocumentBuilder();
            documentBuilder.AddPage(sourceDocument, 1);
            var newDoc = documentBuilder.Build();
            // then create fileStream from newDoc byte[]
```
The source file can be open without any error in Adobe Reader or Chrome. 
But, when I try to open the newly created with Adobe Reader, I encountered this error (It works fine with chrome btw) :

![image](https://user-images.githubusercontent.com/48249095/124279147-6f377b00-db47-11eb-9074-1af0cee7c792.png)

It works fine with many source Pdf but sometimes I'm facing this issue. It depends on the source file 😞 . 
Unfortunatly, due to the sensitive content of these pdf I cannot attached it. 

Have you ever facing such an issue ? 
Let me know if you need more information. 

Thanks in advance, 

Florent

cremor:
I have the same problem with some PDFs that I create with `documentBuilder.AddPage(sourceDocument, pageNumber)`. For me Acrobat Reader shows the error "There was a problem reading this document (14).". 

But I noticed that the problem only happens if I use `documentBuilder.AddPage(sourceDocument, pageNumber)`. When I use `outputDocument.AddPage(sourcePage.Width, sourcePage.Height).CopyFrom(sourcePage)` instead the resulting PDF works fine!

Why are there two different methods to copy a page from a source document? Or better: Why do the two methods behave differently?
@plaisted and @InusualZ seems like you two implemented those two methods?

Some more information that I found for the broken PDFs:

- For now I only get this error for source PDFs that were created by the "Canon SC1011" scanner according to their metadata. Acrobat Reader shows those files without problems before they are merged with PdfPig.
  Sadly, I also can't share the source PDFs since they containt sensitive content.
- Ctrl-clicking on the "OK" button of that error message in Acrobat Reader gives the further information "Expected a dict object."
- Although Acrobat Reader doesn't show the PDF, other programes (e.g. Firefox) can show the PDF fine.
[...]

PR #351, opened 2021-08-02
AddPage fixes for streams and page dict inheritance

plaisted:
Fixes for #344
- Check to make sure objects being converted to direct objects in page dicts aren't streams
- Copy parent MediaBox/CropBox/rotation to child on copy.


EliotJones (approved):
LGTM

PR #358, opened 2021-08-17
\r only in token scanner

Poltuu:
An edge case was lost with this commit https://github.com/UglyToad/PdfPig/commit/31ca3640d2c161e3635187bc1bd94158214a6452?branch=31ca3640d2c161e3635187bc1bd94158214a6452&diff=split

when scanner is only followed by \r (without \n)

### DIFF
--- a/src/UglyToad.PdfPig/Tokenization/Scanner/PdfTokenScanner.cs
+++ b/src/UglyToad.PdfPig/Tokenization/Scanner/PdfTokenScanner.cs
@@ -262,6 +262,7 @@
             }
 
             // From the specification: The stream operator should be followed by \r\n or \n, not just \r.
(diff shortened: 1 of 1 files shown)


InusualZ:
LGTM

I am genuinely interested in knowing who is the producer of the pdf that you encountered with this problem. Because, this is a clear violation of the spec.

EliotJones (approved):
LGTM, will check how this got misplaced.

Poltuu:
@InusualZ Sure. I only got this to give you `/Producer (QZP Version 0.06)` from the pdf.

The only talk about this seems to be this forum post (http://www.quadraforum.fr/viewtopic.php?t=1308).
It must be a specific pdf producer for the HR company "quadra", or "quadratus".

I'm only dealing with payslips, and there is A LOT of weird producers out there.

Issue #354, opened 2021-08-05
PdfDocumentFormatException on GetPage() call


cremor:
I have a PDF that throws a `PdfDocumentFormatException` on a `GetPage()` call. This happens both when opening the PDF with default settings and with `LenientParsingOff`.

```
UglyToad.PdfPig.Core.PdfDocumentFormatException
  HResult=0x80131500
  Message=Invalid type of toUnicode CMap encountered. Got: 340 0.
  Source=UglyToad.PdfPig
  StackTrace:
   at UglyToad.PdfPig.PdfFonts.Parser.Handlers.Type0FontHandler.Generate(DictionaryToken dictionary) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\PdfFonts\Parser\Handlers\Type0FontHandler.cs:line 82
   at UglyToad.PdfPig.PdfFonts.FontFactory.Get(DictionaryToken dictionary) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\PdfFonts\FontFactory.cs:line 44
   at UglyToad.PdfPig.Content.ResourceStore.LoadFontDictionary(DictionaryToken fontDictionary) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\Content\ResourceStore.cs:line 155
   at UglyToad.PdfPig.Content.ResourceStore.LoadResourceDictionary(DictionaryToken resourceDictionary) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\Content\ResourceStore.cs:line 46
   at UglyToad.PdfPig.Parser.PageFactory.Create(Int32 number, DictionaryToken dictionary, PageTreeMembers pageTreeMembers, Boolean clipPaths) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\Parser\PageFactory.cs:line 66
   at UglyToad.PdfPig.Content.Pages.GetPage(Int32 pageNumber, Boolean clipPaths) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\Content\Pages.cs:line 66
   at UglyToad.PdfPig.PdfDocument.GetPage(Int32 pageNumber) in C:\Daten\Projekte\PdfPig-master\src\UglyToad.PdfPig\PdfDocument.cs:line 169
```

Before that two `NullReferenceException` are thrown with the following call stacks, but both are handled by PdfPig:
```
[...]

PR #359, opened 2021-08-22
link annotation fix for PdfDocumentBuilder

plaisted:
- strip link annotations to avoid corrupt links to non-existent pages
- fix issue with acrobat not liking same content stream on multiple pages


cremor:
I can confirm that this PR fixes both previously reported errors in my second PDF in #344. Also, all annotations are now shown correctly. Thanks!

But this PR does not fix #354. That bug already happens with the source PDF, so it would need a change in the reading/parsing code.

plaisted:
@cremor I think the PDF in #354 also triggered a bug when writing the PDF out (maybe you gave sample in #344?) that this should fix, but yes, it won't help with the reading / parsing part, just allow you to write the PDF out without corruption.

EliotJones (approved):
Thanks for the fixes here

cremor:
@EliotJones Would it be possible to merge this in the next few days?
And in case the nightly build still isn't working by then, could you then please also push an alpha package to NuGet?

plaisted:
Looks like something weird with the build, it never ran for PR so can't merge, there a way to manually trigger the build / required check?

Issue #395, opened 2021-12-03
Adobe Reader cannot open files created with pdfPig, and when opening it on chrome, a transparent png is corrupted


mjolivet-lucca:
Hello,

I use PdfPig to open multipages PDFs, split them and recreate them.

Here is a sample of the code :
```csharp
    var stream = File.OpenRead(pdfPath);
    var sourceDocument = PdfDocument.Open(stream);
    var documentBuilder = new PdfDocumentBuilder();
    documentBuilder.AddPage(sourceDocument, 1);
    var newDoc = documentBuilder.Build();
    // then create fileStream from newDoc byte[]
```

It usually works fine, but we have some files which cannot be opened after the split operation with Adobe reader :

![acrobat_error](https://user-images.githubusercontent.com/76952659/144614430-716cb069-0d92-4739-9b7d-99f19c06f89b.png)

translation : 
 > "There was an error processing a page. There was a problem reading this document (110)."

The thing is that when I try to open it in chrome, it works fine, only with a slight issue : the PDF  contains a transparent PNG and in the split file it seems to be corrupted : 

### Original image ###

![edisource_ok](https://user-images.githubusercontent.com/76952659/144614737-69fd6f83-84fa-4715-8e17-65b7d8518ad7.png)

### Split file image ###
 
![edisource_error](https://user-images.githubusercontent.com/76952659/144614714-ae61b256-c5f8-4a4c-a937-30786f5d8616.png)

I can't send you the file since it contains sensitive data (**Edit : I have since succeeded in creating another file which allows us to reproduce the behaviour; It can be found at the end**), but I may have a clue : 
I saw this issue from another pdf library : 
https://github.com/parallax/jsPDF/issues/862

They had something similar happening, and it seems that it was because something was missing with the PNG predictors : 

### jsPDF fix ###

```javascript
getPredictorFromCompression = function getPredictorFromCompression(compression) {
        var predictor;
        switch (compression) {
            case jsPDFAPI.image_compression.FAST:
                predictor = 11;
                break;
            case jsPDFAPI.image_compression.MEDIUM:
                predictor = 13;
                break;
            case jsPDFAPI.image_compression.SLOW:
                predictor = 14;
                break;
            default:
                predictor = 12;
                break;
        }
        return predictor;
    },
```

~~I didn't succeed to recreate another PDF with the same issue.~~ => There is a file to recreate the issue at the end of this comment.

Here are some informations about the file found with PDF Architect : 

[...]

PR #401, opened 2021-12-22
adjust string serialization to handle raw byte data properly

plaisted:
This should resolve issues (or some of them) from #395.

Existing solution was not properly handling non-Utf16BE string. It would potentially escape some data as if it were non-unicode but then convert to unicode after escaping under certain circumstances. This implementation will properly write out with escaping non-unicode pdf string literals.

### DIFF
--- a/src/UglyToad.PdfPig/Writer/TokenWriter.cs
+++ b/src/UglyToad.PdfPig/Writer/TokenWriter.cs
@@ -306,7 +306,8 @@
                 if (pair.Value == null)
                 {
                     WriteToken(NullToken.Instance, outputStream);
-                } else
+                }
+                else
                 {
                     WriteToken(pair.Value, outputStream);
                 }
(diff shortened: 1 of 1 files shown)


mjolivet-lucca (approved):
I tested it and it resolves issue #395 (tested it on multiple files). The code looks good to me.

EliotJones (approved):
Thanks for the fix

src/UglyToad.PdfPig/Writer/TokenWriter.cs
```diff
@@ -407,40 +408,70 @@ private static void WriteStream(StreamToken streamToken, Stream outputStream)
...
+                // happen during parsing as well -> switch to unicode
+                if (stringToken.Data.Any(x => x > 255))
+                {
+                    var data = new StringToken(stringToken.Data, StringToken.Encoding.Utf16BE).GetBytes();
+                    outputStream.Write(data, 0, data.Length);
+                }
+                else
+                {
+                    int ei;
+                    for (var i = 0; i < stringToken.Data.Length; i++)
```
Poltuu, line 438: did you consider putting this code in the `GetBytes()` method of the `StringToken` class, under the `Iso88591` case in the switch on encodindWith ? I feel like that could be a bug, but I could be wrong.

src/UglyToad.PdfPig/Writer/TokenWriter.cs
```diff
@@ -407,40 +408,70 @@ private static void WriteStream(StreamToken streamToken, Stream outputStream)
...
+                        }
+                        else if (c < 32 || c > 126) // non printable
+                        {
+                            var b3 = c / 64;
+                            var b2 = (c - b3 * 64) / 8;
+                            var b1 = c % 8;
+                            outputStream.WriteByte((byte)'\\');
+                            outputStream.WriteByte((byte)(b3 + '0'));
+                            outputStream.WriteByte((byte)(b2 + '0'));
+                            outputStream.WriteByte((byte)(b1 + '0'));
```
Poltuu, line 459: This could be a good use cas for bit manipulation, which should have better performances
```c#
//in this scenario, we can use | in place of +, as ZeroChar equals 0b_0011_0000, and the values are guaranteed lower than 0b_0000_1111 (line 437 + modulo 8)
outputStream.WriteByte(EscapeChar);
outputStream.WriteByte((byte)((c >> 6) | ZeroChar)); // = (c / 64) + 48
outputStream.WriteByte((byte)(((c >> 3) & 0b_0000_0111) | ZeroChar)); // = (c / 8) % 8 + 48
outputStream.WriteByte((byte)((c & 0b_0000_0111) | ZeroChar)); // = (c % 8) + 48
```

src/UglyToad.PdfPig/Writer/TokenWriter.cs
plaisted: I did but looked like other places were using the GetBytes() method to get the raw byte data that was encoded in the string (not serialized byte output)

src/UglyToad.PdfPig/Writer/TokenWriter.cs
plaisted: Hadn't thought of that but if non-printable characters are serialized a lot may be worth updating with your bit logic. Quick test using BenchmarkDotNet shows improvement from ~8ns to ~5ns per char using random values in the pertinent range (0-31,128-255).

Issue #368, opened 2021-09-20
Support opacity for colors?


theolivenbaum:
Hi and great work with the library!
I'm migrating some code to PdfPig and was wondering if it is possible to add a transparent or semi-transparent color? We used it to render invisible text over an image, and to render semi-transparent bounding boxes.

I saw that [SetNonStrokeColorDeviceRgb](https://github.com/UglyToad/PdfPig/blob/c74ca5fda8a51f3af9cf19c675272f57d3beee60/src/UglyToad.PdfPig/Graphics/Operations/SetNonStrokeColorDeviceRgb.cs) only supports RGB, but would be great to have ARGB support too!

theolivenbaum:
hi @EliotJones, any idea on this one? happy to send a PR if you could give some guidance on where to start!

theolivenbaum:
Hi @EliotJones - no worries! I know how it is :)

Thanks for checking it - perhaps the solution for what I'm looking for is something else than transparency - what I wanted was a way to add "invisible text" on top of an image (think a scanned PDF that we add the results of an OCR model). Do you know if this is part of the spec?

Cheers and thanks again for the great work!

JOT85:
I've implemented this privately, I'd love to contribute this feature if you're open to it @EliotJones?

If so, which public API would you prefer? A direct `SetTextRenderingMode` method which adds a SetTextRenderingMode to the content stream (applies to all future text until called again), an `AddInvisibleText` method to add invisible text and then reset the rendering mode, or both?

EliotJones:
Hi @JOT85 thanks, I think `SetTextRenderingMode` is probably preferable to give direct control to consumers. I'd just maybe include a doc-comment to the effect of "use mode 3 to add invisible text" or similar, maybe the method parameter should be an enum:

```
enum RenderingMode {
   FillText = 0,
   // etc
}
```

there might already be one internally that could be publicly exposed.

PR #420, opened 2022-02-15
Add `SetTextRenderingMode` method to `PdfPageBuilder`

JOT85:
Adds support for setting the text rendering mode when creating PDFs, as discussed in #368.

The tests are somewhat limited unfortunately, since I don't think there's a way of checking parsed text rendering mode with the current API, it only checks that the text is correctly inserted and can be read again. Is adding rendering mode information to letters  (when parsed) something you'd be interested in?


EliotJones (approved):
Thanks for the contribution.

I'm not certain on exposing rendering mode for letters. It would be useful I just don't know if there's another way to do it that would avoid adding even more members to the Letter API, e.g. a utility method to get the mode for a letter from the pdf or similar. Let me think about it.

PR #430, opened 2022-03-11
Fixed null ref exception in PdfDocumentBuilder

zlangner:
Fixed null ref exception in PdfDocumentBuilder that occurs when TokenScanner returns null for the IndirectReferenceToken provided.


EliotJones (approved):
Thanks, if you want to make this non-draft it looks good to me, there are often documents with 'abandoned' refs so this makes sense as a fix.

zlangner:
Thanks @EliotJones I just wanted to open an issue with more detail and a sample file for how I found this issue before opening the PR.

PR #435, opened 2022-03-28
Remove recursive algorithm on netstandard2.0 and above

theolivenbaum:


EliotJones:
Great, thanks for the fix. I ran it locally on the full test set and didn't encounter any issues.

I'm not sure if the `#if netstandard` check region is needed, I removed it in this patch and it still appears to build fine, was this needed because you didn't have full framework locally to your dev machine?

[all-frameworks-catalog-factory.txt](https://github.com/UglyToad/PdfPig/files/8403110/all-frameworks-catalog-factory.txt)

theolivenbaum:
That was probably the case! Was complaining that the old net framework didn't have a stack I think...

EliotJones:
Ah looks like CI is running a different version of something to me, I'll try moving some stuff around to see if I can make it build without the directive.

Issue #433, opened 2022-03-19
NET6.0 nuGet package


Raffaello:
It would be possible to create a NET6.0 package too as now only 4.5 and 2.0 are available?

Raffaello:
I have tried just now to compile for net6.0 and apart few tests failing it looks ok.
Can't check if those test where failing also in other version of .NET, but i suppose they weren't.

But the errors looks just around "encryption/decryption"

the tests failing are just:
- Can decrypt date string: 
```
Expected: D:20180808103317-07'00'
Actual:   D:20180808103317
``` 
- GetCorrectBBoxLinux, this look ok to me, it is just a matter of numeric precision
```
Expected: ···759.48), 2.495859375, 0]
Actual:   ···759.48), 2.495859375000002, 0]
```
- EncryptedDocument: CanOpenDocumentAndGetPage (can't read TrueType Tag)
```
PdfDocument.GetPage(Int32 pageNumber) line 166
PdfDocument.GetPages()+MoveNext() line 180
EncryptedDocumentTests.CanOpenDocumentAndGetPage() line 29
----- Inner Stack Trace -----
TrueTypeDataBytes.ReadTag() line 88
TrueTypeFontParser.ReadTable(TrueTypeDataBytes data) line 46
TrueTypeFontParser.Parse(TrueTypeDataBytes data) line 31
CidFontFactory.ReadDescriptorFile(FontDescriptor descriptor) line 119
CidFontFactory.Generate(DictionaryToken dictionary) line 53
Type0FontHandler.ParseDescendant(DictionaryToken dictionary) line 135
Type0FontHandler.Generate(DictionaryToken dictionary) line 53
FontFactory.Get(DictionaryToken dictionary) line 44
ResourceStore.LoadFontDictionary(DictionaryToken fontDictionary) line 155
ResourceStore.LoadResourceDictionary(DictionaryToken resourceDictionary) line 46
PageFactory.Create(Int32 number, DictionaryToken dictionary, PageTreeMembers pageTreeMembers, Boolean clipPaths) line 66
Pages.GetPage(Int32 pageNumber, Boolean clipPaths) line 66
PdfDocument.GetPage(Int32 pageNumber) line 160

```

- EncryptedDocument: Can Provide Multiple Passwords ( same as above)

EliotJones:
Closing this as resolved, the .NET 6 versions are now available in the 0.1.6 previews.

PR #436, opened 2022-04-03
Net6 target framework

EliotJones:
#433 includes details of issues when targetting .NET 6, this adds support for .NET 6 by fixing identified breaking changes in the AES API https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/partial-byte-reads-in-streams

Adds .NET 6 target framework and runs tests on .NET 6.


EliotJones:
Warning, obsolete https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0022

Issue #381, opened 2021-10-20
Stack overflow


theolivenbaum:
Similar to #33 , I'm getting a stack overflow on the method UglyToad.PdfPig.Parser.CatalogFactory.ProcessPagesNode

The stack trace I managed to capture from our logs is here:

`````
 Stack overflow.
    at System.Collections.Generic.Dictionary`2[[UglyToad.PdfPig.Core.IndirectReference, UglyToad.PdfPig.Core, Version=0.1.5.0, Culture=neutral, PublicKeyToken=605d367334e74123],[System.__Canon, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].FindValue(UglyToad.PdfPig.Core.IndirectReference)
    at System.Collections.Generic.Dictionary`2[[UglyToad.PdfPig.Core.IndirectReference, UglyToad.PdfPig.Core, Version=0.1.5.0, Culture=neutral, PublicKeyToken=605d367334e74123],[System.__Canon, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].TryGetValue(UglyToad.PdfPig.Core.IndirectReference, System.__Canon ByRef)
    at UglyToad.PdfPig.Tokenization.Scanner.ObjectLocationProvider.TryGetCached(UglyToad.PdfPig.Core.IndirectReference, UglyToad.PdfPig.Tokens.ObjectToken ByRef)
    at UglyToad.PdfPig.Tokenization.Scanner.PdfTokenScanner.Get(UglyToad.PdfPig.Core.IndirectReference)
    at UglyToad.PdfPig.Parser.Parts.DirectObjectFinder.TryGet[[System.__Canon, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](UglyToad.PdfPig.Tokens.IToken, UglyToad.PdfPig.Tokenization.Scanner.IPdfTokenScanner, System.__Canon ByRef)
    at UglyToad.PdfPig.Parser.CatalogFactory.ProcessPagesNode(UglyToad.PdfPig.Core.IndirectReference, UglyToad.PdfPig.Tokens.DictionaryToken, UglyToad.PdfPig.Core.IndirectReference, Boolean, UglyToad.PdfPig.Tokenization.Scanner.IPdfTokenScanner, Boolean, Int32 ByRef)

....
[...]

PR #473, opened 2022-07-29
Fix page number order.

grinay:
For case when root nod has reference to page the order will be incorrect.
The case if root node has reference
[2 0 R 3 0 R 10 0 R]
Where 2 0 R is intermediate node containing page 1,2,3

Where 3 0 R is intermediate node containing page 4,5,6
Where 10 0 R is page 7

without that fix 7 page will in array as page 1


grinay:
Fix isn't correct. Still has issue with certain documents, will do another fix.

EliotJones:
Thanks for your PR I'll try and get some time to review it soon. My one worry is if it works still for this previously reported issue: https://github.com/UglyToad/PdfPig/issues/381 I don't think there was ever a test case for it but I'll double check during the review

EliotJones (approved):

PR #492, opened 2022-09-15
add option to strip annotation

EnraH:
The new flag enables to strip all ennotations from a page, when adding to a new DocumentBuilder.


EliotJones (approved):

EliotJones:
I can no longer remember why this went unmerged so I'm going to merge it and see what happens. I'll probably wrap the boolean option in an options object so it will be easier to extend in future

PR #504, opened 2022-11-04
Add Test case to support \r\n in EndText

Ali-YousefiTelori:
Hi,
There is a bug in getting the text of my pdf example, Some text does not extract perfectly.

extracted text:
"A slightly more complicated solution is necessary is s.Length exceeds Int32.MaxValue. But ifyou need to read a stream that large into memory, you might want to think about a differentapproach to your problem.This is a text example from Ali Yousefi. Everybody knows this is a test example!"

1.differentapproach => different approach
2.problem.This => problem. This or have a \r\n
3.ifyou => if you

I think this is mixed with every line in the pdf.
I created a test case for you, Hope you find this bug easy.


EliotJones:
Hi @Ali-YousefiTelori thanks for your proposed text. In this case the logic of `page.Text` is behaving as expected. Since `page.Text` is the raw content stream of the document without any additional logic it's usual / expected that words run together, are out of order, are missing whitespace, etc.

The solution here is to use one of the built in word extractors to get words and then use them as needed, e.g. `var words = mainPage.GetWords(NearestNeighbourWordExtractor.Instance);`

https://github.com/UglyToad/PdfPig/wiki/Document-Layout-Analysis

More explanation here: https://dev.to/eliotjones/reading-a-pdf-in-c-on-net-core-43ef

I'll close this for now but let me know if you disagree

PR #515, opened 2022-11-24
Add support for text colors with alpha (aka transparency)

fnatzke:
Have a pdf document that uses gs operator to change the alpha state of non-stroked graphics to opaque (hidden) so show-text (operator Tj) is not shown  but then writes the text using path operators. The two are in similar places but render differently. 

This change 
1. In BaseDrawngProcessor.cs update SetNamedGraphicsState() to supports the 'ca' and 'CA' tokens in the gs operator to set 
     'AlphaConstantStroking' (renamed from just 'AlphaConstant' and a new matching 
     'AlphaConstantNonStroking'.
2. Adds a new class AlphaColor (supporting IColor interface) but can wrap any existing color (RGBColor, CMYKColor, GrayColor,...) but add alpha constant (also decimal).
3. In System.Drawing updates SystemDrawingExntions.cs method ToSystemColor() to check for new AlphaColor and return a system color which uses not only rgb but also alpha.
4. In BaseDrawngProcessor.cs update ShowText() to use AlphaColor to pass down DrawLetter.

### DIFF
--- a/src/UglyToad.PdfPig.SystemDrawing/SystemDrawingExtensions.cs
+++ b/src/UglyToad.PdfPig.SystemDrawing/SystemDrawingExtensions.cs
@@ -60,9 +60,15 @@ namespace UglyToad.PdfPig.SystemDrawing
             if (pdfColor != null)
             {
                 var colorRgb = pdfColor.ToRGBValues();
+                if (pdfColor is AlphaColor alphaColor)
+                {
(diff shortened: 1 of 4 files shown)


BobLd:
Looks good to me, thanks for the PR (+ clear explanations). Reviewed the code but did not test as it's only a merge into the `pdf-to-image` branch

PR #520, opened 2022-12-08
Adding non regression unit test to cover PR 473

mjolivet-lucca:
we had an issue when using v 0.1.6 with some PDF Files using indexed page table.

This issue was resolved with this PR : https://github.com/UglyToad/PdfPig/pull/473

I add a unit test to cover this case and be sure to always have the good order on those specific files.


Poltuu (approved):
thanks for the extra test 👍

EliotJones (approved):
Thanks for the test 🎉

PR #553, opened 2023-02-16
Add enums for PDF/A-3 a and PDF/A-3 b conformance + optional parameter to set out builder parameters in PdfMerger

DGdev91:
https://en.wikipedia.org/wiki/PDF/A
PDF/A-2 and PFD A-3 are almost identical, if not for the embedded files support added in the A-3 standard.
A PDF created using the A-2 rules (without any embedding) should then be already compliant also with the A-3 standard, and there shouldn't be any reason to not allow PdfPig to set 3A conformance in the XMP.


EliotJones (approved):
Thanks for the contribution 🎉

PR #555, opened 2023-02-25
Create BaseStreamProcessor<T> and update ContentStreamProcessor to implement BaseStreamProcessor

BobLd:
First step in order to implement image renderer:
- Create `BaseStreamProcessor` as abstract class and put most logic here
- ContentStreamProcessor now only implement logic to create `PageContent`

No change expected in logic

Part of the project https://github.com/UglyToad/PdfPig/projects/5#card-88238537

POC available in branch [pdf-to-image-skia-color-space](https://github.com/UglyToad/PdfPig/tree/pdf-to-image-skia-color-space)

### DIFF
--- a/src/UglyToad.PdfPig.Tests/PublicApiScannerTests.cs
+++ b/src/UglyToad.PdfPig.Tests/PublicApiScannerTests.cs
@@ -92,12 +92,15 @@
                 "UglyToad.PdfPig.Filters.IFilterProvider",
                 "UglyToad.PdfPig.Functions.FunctionTypes",
                 "UglyToad.PdfPig.Functions.PdfFunction",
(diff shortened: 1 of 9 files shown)


mvantzet:
The changes to BaseStreamProcessor conflict with the changes in my PR #561 regarding cropped/rotated pages so one of us will have to redo some work :-) Other than that LGTM.

I see that IFont will become a public interface, which is nice as it is useful to have more information about fonts. I have some code to get general information (ascender, cap height, descender, italic angle, x height) from the various metrics/descriptors so we can add that to the FontDetails (will issue another PR for that).

BobLd:
@mvantzet great news about the Font and general information you're going to add, that will be useful.

And now worries, I think yours will go first because it's smaller, I'll rework mine thereafter

Issue #550, opened 2023-02-15
Adding a specific PNG to a PDF causes the PDF to be invalid


cremor:
If you add [this](https://user-images.githubusercontent.com/116314/219009895-937bd4a8-9969-4be0-b354-a41919917d9f.png) PNG with `PdfPageBuilder.AddPng()` to a PDF then the resulting PDF can't be shown in Adobe Acrobat Reader. Adobe then shows an error message box about a graphics error and the page is empty.

fnatzke:
@cremor, Can reproduce your issue.  The Adoube Acrobat Reader pop-up message I get is: "A drawing error occurred.". Opens and renders fine in Chrome (without error). Looking further tomorrow.

cremor:
Thanks for your analysis. I think it's not just a bug in Adobe Acrobat Reader because the application Wondershare PDFelement also behaves weird with it. While it can show the image in the PDF initially, it will have various rendering errors in the PDF after you edit the file with Wondershare PDFelement.

That could also be a bug in Wondershare PDFelement, but I think the chance that the PDF is malformed is higher since two completely separate applications have problems with it.

fnatzke:
I'll keep looking. For Adobe I couldn’t find a log file , kb article or hint to the reason. There isn’t anything obvious that stands out in the PdfPig code. I’ll give Wondershare a go. I’m also get a collection of (different) PNGs to try to see if can reproduce to see if there is a common properties that cause.

cremor:
Another hint: If you open the PNG in Paint and just resave it as PNG then the bug doesn't happen. I don't know enough about PNGs to check for differences, but it seems like the way the original file was saved differes from the way that Paint saves PNGs.

fnatzke:
That is interesting. Paint usually saves withOUT the alpha channel. I had that on my todo (mod lib code to drop alpha (smash) as a test) but I’ll put as the next thing to try. Thank you.

fnatzke:
Have what looks to be a fix (still pending full test suite).

In src/src/UglyToad.PdfPig/Writer/PdfPageBuilder.c Line 686:

https://github.com/UglyToad/PdfPig/blob/master/src/UglyToad.PdfPig/Writer/PdfPageBuilder.cs#L686


Change  "png.Header.BitDepth" 
       {NameToken.BitsPerComponent, new NumericToken(png.Header.BitDepth)}, 
to "8"
       {NameToken.BitsPerComponent, new NumericToken(8)},  


[...]

PR #556, opened 2023-03-01
Fix for issue #550

fnatzke:
See issues [discussion](https://github.com/UglyToad/PdfPig/issues/550#issuecomment-1435479121).

BitsPerComponent should be 8 after the given (16-bit) PNG image is replaced 8-bit.

Geting (8 byte) pixel stream should convert each R,G,B value from 2 bytes to 1 byte (8 bit) values.

BitsPerComponent for SMask should be 1.

Test cases try to cover 8 and 16 rgb (a),  grayscale and some other variations.


EliotJones (approved):
Thanks for the fix

PR #570, opened 2023-03-19
AlphaConstants and more named graphics state

BobLd:
Update CurrentGraphicsState with `AlphaConstantNonStroking` and `AlphaConstantStroking` and implement more named graphics state

Previous commited in a the `pdf to image rendering` branch by @fnatzke, see https://github.com/UglyToad/PdfPig/pull/515

I've kept the commit small, doing the straightforward named tokens only and not adding `AlphaColor`. `AlphaColor` is not a pdf concept, so it's up for discussion to add it or not in an other PR


EliotJones (approved):
So what is alpha, I'm very out of the loop but I'm guessing it's colorspace related?

BobLd:
@EliotJones Alpha is just transparency, pdf sets it like color. There's one transparency variable for stroking, and one for non-stroking. Using alpha, you can have shape, text, etc. that's not completely opaque.

BobLd:
looking again at the code, we should use `else if` and not `ifs`, let me update that

PR #573, opened 2023-03-21
Fixes #567, ReadShort now casts to and returns an ushort

### LINKED ISSUE #567: AddJpeg fails, JpegInformation has wrong values
This issue is similar to #394, except that the final result is a white image (and that my JPEG file has no EXIF data).
Image size is 6369x4724, however JpegInformation result is -15813x26036 with BitsPerComponent being 59.

Unfortunately I cannot share the image but looking at the file I can see the header seems to be extremely big to a point where the size information is at byte 572121, the **7th** and last instance of 0xFFC0 maker.

GiovanniNova:
Simply replaces short with ushort allowing values greater than 32767 to work correctly as it should.
Perhaps renaming ReadShort to ReadUShort is also necessary?

### DIFF
--- a/src/UglyToad.PdfPig/Images/JpegHandler.cs
+++ b/src/UglyToad.PdfPig/Images/JpegHandler.cs
@@ -129,7 +129,7 @@
             throw new InvalidOperationException();
         }
 
-        private static short ReadShort(Stream stream, byte[] buffer)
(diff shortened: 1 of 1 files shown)


EliotJones (approved):
Thanks for the fix

PR #552, opened 2023-02-16
Support adding outline bookmarks to existing pdf document

yufeih:
- Adds the ability to create and modify outlines of a PDF document through the `PdfDocumentBuilder` interface.
- Adds a new `UriBookmarkNode` type to represent the `URI` action, allow outlines to point to URLs on the web, that supports both read and write.
- Writing `ExternalBookmarkNode` is not implemented because the current interface lacks the required `/D` key that specifies the destination in the external file to jump to, and we don't seem to need the feature.

Fixes https://github.com/UglyToad/PdfPig/issues/156

Once this is shipped, we can switch from `iTextSharp` to `PdfPig`: https://github.com/dotnet/docfx/issues/4250

PR #579, opened 2023-03-24
Annotations named destinations

mvantzet:
Retracted previous PR, made some changes after conflicts due to PR #552. 

I decided to bite the bullet and add Actions, which works well with the new code. I have some ideas to further clean up Bookmarks code to work with Actions; decided to pick that up later otherwise this PR becomes too big.


BobLd:
@mvantzet are you done working on that? Will review when good

mvantzet:
Yes I am done

BobLd:
I realised there's a lot of changes in code that was written by @EliotJones, maybe it's better I leave it to him.

Just a side note in term of naming convention, do we want to use the name `Action` for the pdf actions, at the risk of confusion with `System.Action`, or should we call them something like `PdfAction`? No strong opinion on that though

mvantzet:
@EliotJones Can you review this PR? Let me know if you want to change Action to PdfAction as @BobLd suggested.

EliotJones:
Hi @mvantzet I'll try to take a look over the Easter break. I think renaming `Action` to `PdfAction` makes sense since `Action` is in `System` iirc so it'll always be present in files when working with this library and therefore namespace separation doesn't really work well. It will also match the convention of e.g. `PdfRectangle` and `PdfPoint`.

mvantzet:
If you could take a look at it that would be great. I will change Action to PdfAction. Is it ok to leave (e.g.) derived class GoToAction the same, or do you want that changed to PdfGoToAction?

EliotJones:
> If you could take a look at it that would be great. I will change Action to PdfAction. Is it ok to leave (e.g.) derived class GoToAction the same, or do you want that changed to PdfGoToAction?

Thanks. `GoToAction` is fine, I think it's just the top level `Action` class that would cause the auto-complete confusion.

src/UglyToad.PdfPig/Actions/ActionProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+    using Tokens;
+    using Outline.Destinations;
+    using Util;
+
+    internal static class ActionProvider
+    {
+        /// <summary>
+        /// Get an action (A) from dictionary. If GoTo, GoToR or GoToE, also fetches the action destination.
+        /// </summary>
+        /// <param name="dictionary"></param>
```
EliotJones, line 16: Imo it's better to remove these empty `<param` items from the doccomments than leave them since it just adds to visual noise without any benefit

EliotJones (approved):
Thanks for the work here

BobLd:
@mvantzet some integrations tests are failing, see https://github.com/UglyToad/PdfPig/actions/runs/4659111370

Maybe have a look if you have time. Must be related to C# versions I think. I've created a PR that might also fix it #597

PR #582, opened 2023-03-25
Setting C# version to latest in Examples and ConsoleRunner projects

BobLd:
Might be related to CI issue with C# version

### DIFF
--- a/examples/UglyToad.Examples.csproj
+++ b/examples/UglyToad.Examples.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <LangVersion>latest</LangVersion>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>
--- a/tools/UglyToad.PdfPig.ConsoleRunner/UglyToad.PdfPig.ConsoleRunner.csproj
+++ b/tools/UglyToad.PdfPig.ConsoleRunner/UglyToad.PdfPig.ConsoleRunner.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <LangVersion>latest</LangVersion>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
   </PropertyGroup>


EliotJones (approved):

cremor:
Why not simply update the target framework to a current (and still supported) version? That would increase the C# version automatically.

EliotJones:
> 100% agree with @cremor, might be good to update to dotnet 6. Any opinion @EliotJones? Happy to create the PR if need be

So the reason for the current target frameworks being excessively broad is based on my past experience in .NET shops. A lot of these places are going to be running ancient versions and the migration paths off them are going to be long too, in the multiple years to bump the version. Primarily I was working with government software that I'd be surprised if some instances have even reach NET 4 yet.

The only reason I haven't gone back further to .NET 3 versions is it gets painful to write in versions that early but by keeping the library dependency free, with the exception of valuetuple and being careful with syntax means the library will continue to support the unfortunate Devs maintaining these legacy apps.

I'm in favour of adding newer version support too though.

All that being said I'm not actually sure what of the past like 7 C# versions are restricted to which frameworks versus being dictated by language version versus ci runner. Effectively as long as the ci can run tests for 1 legacy framework (probably 4.8) and 1 'core' version (6 or 7) and we can still build all target framework versions for publishing I'm in favour of being as close to the latest language features as possible.

We can maybe retire some of the earlier core versions since you'd assume users on core have a quicker upgrade path so the support matrix might be:

- NET Framework 4.5+
- NET (Core) 3+
Let me know if you disagree though and we can probably retire some older full framework versions

cremor:
I didn't mean to suggest updating the target framework for the library itself. Only for the projects that were updated in this PR, as an alternative to setting `LangVersion`. And those can target net7.0 in my opinion.

If you want to change the target frameworks of the library itself, I'd suggest to follow Micrsoft's guideline: https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting
Basically:
- netstandard2.0
- net461
- (Older .NET Framework versions if you want to support those old versions which are not supported by Microsoft any more)
- net5.0/net6.0/net7.0 only if those would offer additional features compared to netstandard2.0

BobLd:
@EliotJones As mentionned above, I think changing the target frameworks of the library to only `netstandard2.0`, `net461` and `net6.0` makes sense.

We can update Examples and ConsoleRunner projects to `net6` too

PR #583, opened 2023-03-25
Move IColorSpaceContext to CurrentGraphicsState

BobLd:
As per "Pdf Reference 1.7", section 8.4 Graphics State, Table 52, the color space should be part of the graphics state (together with the CTM, etc.)
image

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Graphics/TestOperationContext.cs
+++ b/src/UglyToad.PdfPig.Tests/Graphics/TestOperationContext.cs
@@ -26,15 +26,13 @@
 
         public PdfPath CurrentPath { get; set; }
 
-        public IColorSpaceContext ColorSpaceContext { get; }
-
         public PdfPoint CurrentPosition { get; set; }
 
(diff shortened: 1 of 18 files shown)


BobLd:
As an FYI, I think this was never an issue in the past as the `SetNonStrokeColor()` and `SetStrokeColor()` _guess_ the color space from the number of operands passed (see below) - which I believe is not the correct way to do but worked very well so far. The actual current color space (wrong one) was ignored and overridden with the more correct one base on the number of operands.

This will become an issue when we actually implement all color spaces, as we won’t be able to do guessing.

I noticed this was happening in the 68-1990-01_A.pdf document on page 7

https://github.com/UglyToad/PdfPig/blob/58b3394d01d71a9cc5be71d82509839f437e2c28/src/UglyToad.PdfPig/Graphics/Operations/SetNonStrokeColor.cs#L36-L51

BobLd:
Closing this one as I'll create a bigger PR to implement the color spaces

PR #595, opened 2023-04-10
Properly implement color spaces

BobLd:
- Use ColorSpaceDetails instead of ColorSpace enum
- Move color spaces logic inside respective color space details
- Actually use separation color space
- #583 
- ResourceStore: Store loaded name color spaces and make namedColorSpaces a StackDictionary (added tests)

I don't want to make the PR bigger as I think there's enough to review. I will implement the missing color spaces in coming PRs

This PR follows the way PdfBox handles color spaces


src/UglyToad.PdfPig/Content/IPdfImage.cs
```diff
@@ -27,17 +27,6 @@ public interface IPdfImage
...
-        /// <summary>
-        /// The <see cref="ColorSpace"/> used to interpret the image.
-        /// This defines the number of color components per sample, e.g.
-        /// 1 component for <see cref="Graphics.Colors.ColorSpace.DeviceGray"/>,
-        /// 3 components for <see cref="Graphics.Colors.ColorSpace.DeviceRGB"/>,
-        /// 4 components for <see cref="Graphics.Colors.ColorSpace.DeviceCMYK"/>,
-        /// etc.
-        /// This is not defined where <see cref="IsImageMask"/> is <see langword="true"/> and is optional where the image is JPXEncoded for <see cref="XObjectImage"/>.
-        /// </summary>
-        ColorSpace? ColorSpace { get; }
```
EliotJones, line 39: We'll need to document a migration path for these public API members, if you're able to pop a brief summary in a comment that would be great thanks

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceExtensions.cs
```diff
@@ -36,20 +36,21 @@ public static ColorSpaceFamily GetFamily(this ColorSpace colorSpace)
...
         {
             colorspace = ColorSpace.DeviceGray;
 
-            if (name.Data == NameToken.Devicegray.Data)
+            if (name.Data == NameToken.Devicegray.Data || name.Data == "G")
             {
                 colorspace = ColorSpace.DeviceGray;
             }
-            else if (name.Data == NameToken.Devicergb.Data)
+            else if (name.Data == NameToken.Devicergb.Data || name.Data == "RGB")
```
EliotJones, line 49: Given how few documents actually follow the spec would it make sense to do a case insensitive check in this file, e.g.

```
string.Equals(name.Data, "rgb", StringComparison.OrdinalIgnoreCase)
```

EliotJones (approved):
I've lost track of what's going on here but it looks like the right path to take :D

src/UglyToad.PdfPig/Content/IPdfImage.cs
BobLd: @EliotJones makes sense, I'll post something here

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceExtensions.cs
BobLd: Hehe I agree with you on the idea but it seems we're safe on this as this is how it was done even before the change (see here src/UglyToad.PdfPig/Util/ColorSpaceDetailsParser.cs)

I'll have a look in my next PR

PR #601, opened 2023-04-12
Fix integration tests for #579

BobLd:


BobLd:
@EliotJones just a side note about merging types (not related the this PR in particular):
Instead of using a classic merge in the repos, I use the 'Rebase and merge' or 'Squash and merge' option when doing the merge on github. I allows keeping a linear history and squashing many commits into one when necessary

EliotJones (approved):
Thanks for fixing this, one day we'll get the CI server to use a good language version, if it's possible, I'm guessing we're bottlenecked by .NET 4.5

PR #606, opened 2023-04-14
Revert "Update run_integration_tests.yml"

BobLd:
This reverts commit 92f47df9f6c6ff6d01df9dc8815a17e63ca3409d.

@EliotJones should we trigger the run_integration_tests on PR to master too? like the other tests. Would avoid discovering issues after PR has been merged


EliotJones:
@BobLd the main reason for running only on master is the source archive is on my personal VPS and could generate too much traffic if run on every PR

EliotJones (approved):

BobLd:
okay makes a lot of sense

PR #616, opened 2023-04-22
Set C# LangVersion to 7.3 to align with integration tests

BobLd:
Set the LangVersion to 7.3 to fix integration tests failing

### DIFF
--- a/tools/UglyToad.PdfPig.ConsoleRunner/UglyToad.PdfPig.ConsoleRunner.csproj
+++ b/tools/UglyToad.PdfPig.ConsoleRunner/UglyToad.PdfPig.ConsoleRunner.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>7.3</LangVersion>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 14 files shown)


tools/UglyToad.PdfPig.ConsoleRunner/UglyToad.PdfPig.ConsoleRunner.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>7.3</LangVersion>
```
EliotJones, line 4: I know you've probably already attempted something similar and I didn't understand fully. But is it not possible instead to bump this to latest and change the target framework on line 6 to something like .NET 6 instead?

BobLd: @EliotJones I think it would work too, I'm also confused as to why we have the issue to begin with tbh... 

What I did here is in order to get the same errors in VS, and fixing the error messages.

I think if we increase the dotnet version to 6 here we would also need to increase the dotnet version in the GitHub Action that runs the integration tests.

Really up to you ;)

PR #629, opened 2023-05-22
Fix: copyfrom resources collisions

thinkbeforecoding:
When copying multiple pages from a source document to a single destination page, there is a conflict not only with resources prefixed by I (images), but also Fm (containing drawing instructions) and ExtGfxState (used for opacity etc).

This change resolves conflict during page copyFrom.

I introduced a new class to manage name collision (using a default prefix, but that can also fix existing names by extracting the prefix)

The ExtGfxState has to be handled separately since it is referenced in a different operation and is located in a different dictionary.

There are probably other names/resources/instructions that need to be handled in the same way, like ColorSpaces etc...


EliotJones (approved):
Thanks for the fix, it's a good approach

PR #632, opened 2023-05-26
Use double everywhere instead of decimal

BobLd:
Replacing `decimal` with `double` accross the code base as I don't think we really need to use `decimal` for what we do. I also tried to remove useless casting.

Really not an important PR, but makes the code more efficient. Given the big number of files to reviews versus the actual little improvement, feel free to not merge it @EliotJones

### DIFF
--- a/src/UglyToad.PdfPig.Core/PdfPoint.cs
+++ b/src/UglyToad.PdfPig.Core/PdfPoint.cs
@@ -4,7 +4,7 @@
     using System.Globalization;
 
     /// <summary>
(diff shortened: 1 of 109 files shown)


EliotJones:
Thanks for the changes here @BobLd I think this is a good end state but it would be better to attack it piecemeal since I think there may be some places, e.g. writing values to file, etc, where decimal is a better bet. I think we can change as-we-go.

EliotJones:
I'll close this for now

PR #634, opened 2023-05-28
Implement hflex and hflex1 operators in Type2CharStringParser (fix #558)

### LINKED ISSUE #558: Type 2 font bounding box (and path) wrong
Some of the type 2 font letters in this document have wrong bounding box and path.

This seems to happen with letter that have a `vlineto` operator, tough not sure at all about that.

[TIKA-2121-0.pdf](https://github.com/UglyToad/PdfPig/files/10892296/TIKA-2121-0.pdf)

Below is a render of the paths
image

Bounding boxes are not completly wrong but still a bit off, inline with the render above (below uses 0.1.8-alpha-20230219-999f9):
image

BobLd:
Fix for #558


EliotJones (approved):
Thanks

Issue #647, opened 2023-06-17
Render pdf as images


BobLd:
Hi @EliotJones, below are my initial thoughts about what's missing to enable rendering pdf pages as images:

- Allow users to create their own `ContentStreamProcessor`, either starting from scratch implementing `IOperationContext`, or implementing an abstract class (to be added, let's call it `BaseStreamProcessor<T>` for now) that contains the current main logic of `ContentStreamProcessor`
- This new abstract `BaseStreamProcessor<T>.GetContent(...)` should return any T, be it `PageContent`, `byte[]`, etc
- The image renderer will derive from`BaseStreamProcessor<byte[]>` for example, with `byte[]` being the rendered page image
- `PageFactory` needs to accept any `IOperationContext` implementation, and the `GetContent()` should return `T`
- We might want to use a similar pattern as in PdfBox, creating a `PdfRenderer` helper class that will take the `PdfDocument` in the constructor, and call the page factory.

I think it would be good to make it possible to not reload the `PdfDocument` but to make it possible to not have to get the `Page` object before rendering.

FYI, some notes on how PdfBox does it:
- `PDFRenderer.java`
	- Helper class
	- ctor: `PDFRenderer(PDDocument document)`
	- `renderImage(int pageIndex, float scale)`
	- uses `PageDrawer.java`
	
- `PageDrawer.java`
	- extends PDFGraphicsStreamEngine (equivalent to PdfPig IOperationContext)

Any feedback welcome!

BobLd:
As a side note, it seems the `GetPage()` function is still not thread safe (as experienced in #168), and would be the perfect opportunity to change that to be able to render pages in parallel. 

I haven't investigated that thoroughly but it seems the `ResourceStore` is one cause of the issue

EliotJones:
Hi @BobLd that's fine, you should just be able to use the plain Apache 2 license for that library since this one inherited a bunch of stuff from PDFBox to do with Adobe files etc.

BobLd:
@EliotJones I've created the repos and published the NuGet package, any feedback welcome

https://github.com/BobLd/PdfPig.Rendering.Skia

I'm closing the issue

PR #656, opened 2023-07-01
[POC] Add AddPageFactory() methods and GetPage<T>() to PdfDocument. Make pu…

BobLd:
@EliotJones This is a POC for rendering pdf pages as image (see #647).

This is meant to be a basis for discussion and it will certainly need more refactoring / modification.

The highlight of the PR are the following:
- Add `AddPageFactory()` methods and `GetPage<T>()` to `PdfDocument`.
- Make public `IPageFactory<TPage>`, `PageFactoryBase<TPage>`, `IResourceStore`, `ILookupFilterProvider`, `IPageContentParser`, `IPdfTokenScanner`, `UserSpaceUnit`, `ResourceColorSpace`

I have not added the actual code to render a page as an image, but this PR allows for it. I'll add a minimal example shortly to do so.

See `PageFactoryTests` to understand how to use the changes. In the tests the `GetPage<T>()` returns `SimplePage` and `PageInformation` bit this could be any object (i.e. `Bitmap`, `byte[]`)

I hope that splitting the changes needed like that makes sense.

Have a look when you have time


BobLd:
@EliotJones now completed with page rendering. To render page as image, do the following (you can also check the Skai tests project)
```csharp
using (var document = PdfDocument.Open(Helpers.GetDocumentPath(path)))
{
	document.AddPageFactory<SKPicture>(typeof(SkiaPageFactory));

	for (int p = 1; p <= document.NumberOfPages; p++)
	{
		var page = document.GetPage(p);

		using (var picture = document.GetPage<SKPicture>(p))
		{
			Assert.NotNull(picture);

			using (var fs = new FileStream($"{path}_{p}.png", FileMode.Create))
			using (var image = SKImage.FromPicture(picture, new SKSizeI((int)(page.Width * scale), (int)(page.Height * scale)), SKMatrix.CreateScale(scale, scale)))
			using (SKData d = image.Encode(SKEncodedImageFormat.Png, 100))
			{
				d.SaveTo(fs);
			}
		}
	}
}
```

BobLd:
@EliotJones what do you think about the approach? If you're good to go, I'm happy with breaking down the changes in smaller PRs so that review will be easier

EliotJones:
As usual sorry for the delay here. Let me check I understand the plan correctly from my initial skim.

1. Current implementation of `document.GetPage` is preserved, using content stream processor to return `Page` by default
2. A new way of registering page factories on the document is added
3. Necessary internal types become public to allow these page factory implementations to hook into the content stream processor equivalent logic
4. A new project at the same logical level as DocumentLayoutAnalysis is added that exposes its own page factory based on SkiaSharp, the output of which is the necessary image type to use with SkiaSharp API
5. This could be extended to an implementation on ImageSharp, System.Drawing, etc.

If this is correct then I think this is a good approach. I'm happy with starting to make some internal stuff public because in all honesty I'm not going to have time to go back and change the implementations anyway so making the library more extensible for consumers is probably the best move.

I like that the implementation is kept library independent since my main aim is to keep the image library out of the core package.

BobLd:
yes you got everything right 👍 

On top of that, the new Skia project should be a new NuGet package on its own - as you said, other implementations (ImageSharp, System.Drawing, etc) might be added later in their own projects/Nuget packages and we want to keep image libraries outside of the core package.

I'm very happy you're okay with the approach. I'll start creating PRs shortly, starting by exposing internal classes as public (feel free to do it if you have time)

Issue #672, opened 2023-08-02
Strange issue parsing a PDF with an OpenType font.


readingdancer:
My client has a bunch of PDF documents that are failing to parse and I have narrowed it down to their headings using the Semplicita Pro OpenType font from Adobe.

However, I created a test PDF using the same font and it works fine, I also removed all other content from my client's PDF and changed the heading text, re-saved it and it still fails. So I am able to provide two example documents, one that works and one that fails.

I don't know if anyone will be able to debug this, but it would be greatly appreciated it you can :)

This works:
[Test Doc.pdf](https://github.com/UglyToad/PdfPig/files/12243295/Test.Doc.pdf)

This fails to parse:
[Why does this not work.pdf](https://github.com/UglyToad/PdfPig/files/12243296/Why.does.this.not.work.pdf)

Thank you in advance!

BobLd:
Issue happens in the `TrueTypeFontParser` when parsing the `TrueTypeDataBytes`, since it's an OpenType font. The data does not contain the necessary tables (i.e. 'head', 'hhea'). 

https://github.com/UglyToad/PdfPig/blob/8a82500427ace6d4dcb1b2ea7cb3fdb5e32c765d/src/UglyToad.PdfPig/PdfFonts/Parser/Parts/CidFontFactory.cs#L142C1-L147C26

It seems the font bytes are actually parsable with `CompactFontFormatParser`, but with some errors though.

I've attached the problematic bytes below, if someone wants to have a look too.
[test_unzip.txt](https://github.com/UglyToad/PdfPig/files/12266792/test_unzip.txt)

BobLd:
@readingdancer May I asked which tool you used to create / edit the pdf files? Do you know which tool your client used?

Asking because the following might be related
https://community.adobe.com/t5/adobe-fonts-discussions/opentype-ps-fonts-are-identified-as-type-1-fonts-when-saving-indesign-files-to-pdf/td-p/11852099

readingdancer:
I used the latest version of Adobe Acrobat and just created a standard blank PDF. Added the text and saved the PDF with the default settings. 

I will ask my client and get back to you next week, my guess is they probably used MS Word, but I will confirm when I find out. 

Thank you for looking into this, I really appreciate it.

readingdancer:
Ok so apparently they use an application called Foxit. 

[https://www.foxit.com/pdf-editor/](https://www.foxit.com/pdf-editor/)

They told me they just save with the default settings using that app.

[...]

PR #674, opened 2023-08-05
Address #672 to ignore errors while reading the descriptor file in CidFontFactory

BobLd:
Address #672 to ignore errors while reading the descriptor file in CidFontFactory


EliotJones (approved):
Looks good way to error handle this fallback, thanks

PR #676, opened 2023-08-05
Check and handle circular references when processing XObject forms and fix #671

### LINKED ISSUE #671: StackOverflow Exception
Hello! I am using version 0.1.8
I get a stack overflow error when getting sheets of a document.
Document attached
[bad.pdf](https://github.com/UglyToad/PdfPig/files/12228439/bad.pdf)

BobLd:
Check and handle circular references when processing XObject forms and fix #671


BobLd:
@EliotJones tests are now fixed, let me know what you think about the method to handle that

src/UglyToad.PdfPig/Graphics/ContentStreamProcessor.cs
```diff
@@ -603,6 +603,12 @@ private void ProcessFormXObject(StreamToken formStream)
             // 3. We don't respect clipping currently.
 
             // 4. Paint the objects.
+            if (parsingOptions.UseLenientParsing && xObjectName != null && operations.OfType<InvokeNamedXObject>()?.Any(o => o.Name == xObjectName) == true)
```
EliotJones, line 606: Would this still stack overflow when `parsingOptions.UseLenientParsing` is `false`? Maybe the other branch of this `if` should throw an exception in that case?

BobLd: it would yes,and makes sense to throw an exception. Will update that

BobLd: I need to do further investigation in fact. MOZILLA-3136-0.pdf is flagged as having circular reference now, but was never an issue with that (it seems)

BobLd: I've updated the logic to also check the forms data

EliotJones (approved):

PR #681, opened 2023-08-11
Enabled ClipPaths option to filter letters.

savasozer:
I hope I've implemented correctly.


src/UglyToad.PdfPig/Graphics/ContentStreamProcessor.cs
```diff
@@ -315,7 +315,16 @@ public void ShowText(IInputBytes bytes)
                 var transformedPdfBounds = PerformantRectangleTransformer
                     .Transform(renderingMatrix, textMatrix, transformationMatrix, new PdfRectangle(0, 0, boundingBox.Width, 0));
 
-                      
+                if (parsingOptions.ClipPaths)
+                {
+                    var currentClipping = GetCurrentState().CurrentClippingPath;
+
+                    if (!currentClipping?.Contains(transformedGlyphBounds) ?? true)
```
BobLd, line 322: No need to call currentState `GetCurrentState()` again here, you can reuse the `currentState` variable from line 246

BobLd (changes requested):
Is it possible to add unit tests too? If you cannot, just let me know I'll have a look

src/UglyToad.PdfPig/Graphics/ContentStreamProcessor.cs
BobLd: I think for the corner case where the clipping path would be null, we need to render the text (i.e. we don't skip). You can achieve that using something like 
```csharp
currentState.CurrentClippingPath?.Contains(transformedGlyphBounds) == false
```

BobLd: @savasozer no worries at all! I agree with you, using `Intersect()` makes more sense

savasozer: I couldn't find a PdfPath PdfRectangle Intersection testing method. So I made it myself using your Contains method :)

BobLd: @savasozer thanks a lot for that, that's very helpful.

Maybe 1 remark on your function (I'm away so I could not actually test it):
- There is a (very rare) edge case I think where the clipping path is completely inside the letter's rectangle, leading your function to return _false_ instead of _true_. Same if the clipping path is smaller than the letter and partially inside and outside of the letter. In both cases all letters point would be outside but actually overlap. There might be other edge cases but I could only think about this one.

BobLd: Maybe this link would be useful https://stackoverflow.com/questions/38382777/how-do-i-determine-if-two-polygons-intersect-using-clipper#38382810

yjagota:
@BobLd Hi Bob. Asking the same question as [jerrosenberg](https://github.com/jerrosenberg). Can anything be done to incorporate this change? I can repurpose the submission of [savasozer](https://github.com/savasozer) along with some tests as there are a lot of code changes done after this pull request. Please look into this if possible. Thanks & Regards.

BobLd (approved):

BobLd:
Apologies for following up on the changes after my comments... This branch now has conflicts and cannot be merged as is. @savasozer I know it's been a while, but can you resolve the conflicts?

savasozer:
Hi @BobLd,

Thank you for reaching out regarding the pull request I submitted one year ago. I understand that resolving the conflicts may be challenging given the number of commits that have been added since then.

Unfortunately, I have since transitioned to a new career in the gaming industry, and as a result, I no longer have the necessary development tools and environment (e.g., Visual Studio) set up to address these conflicts effectively.

Given these circumstances, I’m unable to assist with resolving the conflicts in this pull request. However, I appreciate your understanding and wish you all the best with the project.

If there’s anything else I can help with, please let me know.

yjagota:
I'll submit a new pull request along with two tests by the evening.

yjagota:
Hi @BobLd. You can close this pull request. The fresh one is [here](https://github.com/UglyToad/PdfPig/pull/905).

PR #686, opened 2023-08-22
Load system fonts in all supported platform builds

### LINKED ISSUE #685: Support for OS specific system fonts finder in build targeting >= net.60
I'm developing my application in .NET 6.0 and I'm super happy with the library! Many thanks!

Unfortunately I'm facing some issues loading the correct fonts, when running the application on my linux host (as docker). 
```
OS Architecture:       Arm64
OS Description:        Linux 5.15.0-1026-raspi #28-Ubuntu SMP PREEMPT Fri Mar 10 14:28:52 UTC 2023
OS Platform:           LINUX
Process Architecture:  Arm64
Runtime Identifier:    debian.11-arm64
Framework Description: .NET 7.0.10
```

From what I can see, is that the conditional `#if NETSTANDARD2_0` in `SystemFontFinder.cs` is only available with the Net Standard 2.0 build of the library. 

I'm suggesting to change that, so that the OS font loading mechanism will work on all .NET Core supported platforms. Based on https://learn.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks, and the available builds I would change that part to

Source from: https://github.com/UglyToad/PdfPig/blob/4a480ffd7f30b35e42fc9d664c0891a5344bfae4/src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs#L79C19-L79C19
```
// 
#if NETSTANDARD2_0_OR_GREATER OR NET
// Check based on RuntimeInformation
#else if NETFRAMEWORK
// default to .NET Framework
#else
#error Unsupported target build
#endif 
```

As an alternative, the SystemFontFinder logic could somehow be exposed, so that a custom finder can be registered, but that would probably be a bit more work / alignment needed.

michaelschnyder:
Allow OS specific ISystemFontFInder implementations for all build targets above and include .NET Standard 2.0, closes #685

### DIFF
--- a/src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs
+++ b/src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs
@@ -76,7 +76,7 @@ namespace UglyToad.PdfPig.Fonts.SystemFonts
             NameSubstitutes = dict;
 
             ISystemFontLister lister;
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0_OR_GREATER || NET
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 lister = new WindowsSystemFontLister();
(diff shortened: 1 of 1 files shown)


Numpsy:
Probably trivial, but just in case - as long as there is a .NET6+ target - using the ```OperatingSystem.IsXXX``` functions for the platform specific bits *might* be helpful with trimming or AOT when doing single platform builds?

michaelschnyder:
Can you please elaborate, I don't get your point.

With this change, the mentioned methods are now more widely used, given the change in the if. I don't understand how that would affect AOT in a negative way.

BobLd (approved):

michaelschnyder:
thanks for merging, I updated my solution with the latest build and it does work as expected in a .NET Core 6 docker container on Ubuntu(Raspberry Pi)

PR #690, opened 2023-08-26
Properly handle page rotation for crop box and media box and fix #665

### LINKED ISSUE #665: Cropbox is wrong since update to 0.1.8
See attached example, top part of the text is not in the CropBox area of the page. This is a OCRd PDF where the original is kept as background image and the glyphs are set to invisible. 
[SmallCropbox.pdf](https://github.com/UglyToad/PdfPig/files/12192093/SmallCropbox.pdf)

This worked fine in release 0.1.7

BobLd:
- Properly handle page rotation for crop box and media box 
- Fix #665 and add test
- Make `TransformationMatrix` a `readonly struct` and implement `==` and `!=` operators
- Move `GetInitialMatrix(...)` static method in an helper class


src/UglyToad.PdfPig/Graphics/OperationContextHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            if (userSpaceUnit.PointMultiples != 1)
+            {
+                log.Warn("User space unit other than 1 is not implemented");
+            }
+
+            // After rotating around the origin, our points will have negative x/y coordinates.
+            // Fix this by translating them by a certain dx/dy after rotation based on the viewbox.
+            double dx, dy;
+            switch (rotation.Value)
```
EliotJones, line 53: I'm not certain they exist but I think it would be possible to encounter multiples of 90 > 360, maybe switch on rotation.Value / 90 here (and then values 0, 1, 2, 3, 4)?

EliotJones (approved):
Thanks for the fix here, I'm still shaky on matrices but I assume this is applying the necessary rotation once up front rather than every time in the content stream processor as it was before

BobLd:
@EliotJones from what I recall the media box and crop box where not properly transformed... Maybe we need some pdf pig documentation about matrices in general and add more tests about corner cases

PR #691, opened 2023-08-27
Handle SetFlatness operator with no value and fix #682

### LINKED ISSUE #682: Error reading PDF
This PDF when saved to a file (and many more I can provide if needed) returns null for pages, even though it is readable in Acrobat and Firefox and Chrome.

https://www.nflgsis.com/2013/post/02/56162/Gamebook.pdf

Many other PDF's from the same site are readable.  Everything newer than 2013 for example:  https://www.nflgsis.com/2018/reg/02/57587/Gamebook.pdf

snippet that fails....

            using (PdfDocument pdf = PdfDocument.Open(gamebookName))
            {
                var pdfpages = pdf.GetPages();

                var pages = new List<List<string>>();
                foreach (Page page in pdfpages)   <== NULL PAGES object

BobLd:
fix #682


src/UglyToad.PdfPig/Graphics/ReflectionGraphicsStateOperationFactory.cs
```diff
@@ -129,6 +129,10 @@ public IGraphicsStateOperation Create(OperatorToken op, IReadOnlyList<IToken> op
                 case SetColorRenderingIntent.Symbol:
                     return new SetColorRenderingIntent((NameToken)operands[0]);
                 case SetFlatnessTolerance.Symbol:
+                    if (operands.Count == 0)
```
EliotJones, line 132: Wonder if there's a way to change this mapping to use new codegen stuff where the input is basically expected args, fallback behavior, expected output or something, to reduce the number of times we need to fix individual ones to be more fault tolerant, ah well, a nice thought experiment, LGTM

EliotJones (approved):
Long time no see, sorry for the radio silence, it's tough to find the motivation for me atm. Thanks for the fix

PR #701, opened 2023-09-12
Fix for #662: when encountering invalid content, try to continue parsing

### LINKED ISSUE #662: Dealing with invalid content in streams
I have a 180-page PDF file (that I am unable to share) which has the following content stream on page 79:

```
(..valid content...)
1. #IND
1. #IND
1. #QNAN
1. #QNAN
-176.999923 56.999971 cm
BT
1. #QNB
Tz
/F0 29.1667 Tf
1. #IND
1. #IND
Td
(l) Tj
ET
1. #QNAN
1. #QNAN
1. #QNAN
1. #QNAN
1. #QNAN
1. #QNAN
cm
(..valid content...)
```

The issue is the strings like "1.#IND" and "1.#QNAN" which aren't the decimals that are expected for operator "cm". It seems to be an error of a PDF generating application written in C++, ref https://studiofreya.com/cpp/1-inf-1-ind-and-qnan-floating-point-numbers-and-errors/.

Strictly speaking this PDF document is not valid, but it would be nice to at least parse the rest of the pages (eg. with option UseLenientParsing). Options I see:

(a) skip invalid operators, or
(b) insert dummy values so they become valid, or
(c) abort parsing the page content stream altogether

Opening the file in Adobe PDF reader works, but shows an error message when navigating to this page. It does not find text after the invalid section, so that seems to imply option (c) is used.

Opening the file in Nitro PDF works, does not show an error message, and searching for text after the invalid section works, but it cannot be rendered, so that seems to imply either (a) or (b).

I am willing to work on a fix for this, but which approach should we take?

UPDATE: I just noticed issue #467, will check how that was handled.

mvantzet:
if option "useLenientParsing" is in effect. Added a test document that has a corrupted operation in the content stream
and tests to verify that parsing crashes without "useLenientParsing" but continues with "useLenientParsing". Fixes #662


BobLd:
@mvantzet thanks a lot for the PR! 

One remark: Can you address the warnings in your tests "Do not use Assert.True() to check for substrings."

The rest looks good to me so far

mvantzet:
No problem, done!

BobLd (approved):

PR #707, opened 2023-10-01
Make all structs readonly when possible

BobLd:


EliotJones (approved):
does this improve performance or is the benefit mainly code quality and compiler level checks? wonder if we could also start using records but I must admit I've missed a lot of C# changes recently so not sure what should be struct vs class vs record vs record struct

EliotJones (approved):
does this improve performance or is the benefit mainly code quality and compiler level checks? wonder if we could also start using records but I must admit I've missed a lot of C# changes recently so not sure what should be struct vs class vs record vs record struct

BobLd:
@EliotJones it should improve performance slightly (related to `defensive copy` if I understand correctly, I would need to find back the article I found that was dicussing that), and it improves code quality for sure.

I agree on your point about `record` / `record struct`. I was planning to add a benchmark project at some point, this might be helpfull to assess what should be what

Issue #490, opened 2022-09-07
How to add/write custom XMP metadata


eholman:
Is it possible to add or write custom xmp metadata to a (new) PdfDocument(Builder)?

EliotJones:
Not currently

BobLd:
@eholman closing issue as functionality was added by @listm, feel free to reopen if any question

PR #713, opened 2023-10-06
Add support for custom xmp metadata to PdfDocumentBuilder #490

listm:
Add XmpMetadata option to PdfDocumentbuilder as asked in #490 

Can be used as follows:

```csharp
PdfDocumentBuilder pdfDocumentBuilder = new()
{
    ArchiveStandard = PdfAStandard.A2A,
    IncludeDocumentInformation = true,
    XmpMetadata = XDocument.Parse(@"<x:xmpmeta xmlns:x=""adobe:ns:meta/"">
      <rdf:RDF xmlns:rdf=""http://www.w3.org/1999/02/22-rdf-syntax-ns#"">
        <rdf:Description xmlns:dc=""http://purl.org/dc/elements/1.1/"">
          <dc:title>Document title</dc:title>
        </rdf:Description>
      </rdf:RDF>
    </x:xmpmeta>"),
};
```

XMP will only be inserted if a PdfAStandard is chosen. I think this is sufficient for the start.

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Writer/XmpTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Writer/XmpTests.cs
@@ -1,0 +1,108 @@
+﻿namespace UglyToad.PdfPig.Tests.Writer
+{
+    using Integration;
+    using PdfPig.Content;
+    using PdfPig.Writer;
+    using System.Linq;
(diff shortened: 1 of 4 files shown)


BobLd (approved):
Looks good to me, thanks for the good documentation

PR #717, opened 2023-10-22
Add more classes to the Public API

BobLd:
As discussed in #656.

I've done the changes in several commits - that might help the review. I'll squash them when merging.

I've also update `NamedDestinations` namespace to match folder structure. Feel free to let me know if that should not the case.

For the moment, `IPageFactory` stays internal, mainly because `InternalParsingOptions` is. Same for `IResourceStore` (and related `XObjectFactory`).

@EliotJones what's your view on `InternalParsingOptions`? I don't want to make it public but we will need to find a way to make `IResourceStore` and `IResourceStore` public


src/UglyToad.PdfPig/Content/PageTreeMembers.cs
```diff
@@ -6,17 +6,26 @@
...
-        public CropBox GetCropBox()
+        internal CropBox GetCropBox()
         {
             return null;
         }
 
+        /// <summary>
+        /// The page media box.
+        /// </summary>
         public MediaBox MediaBox { get; set; }
```
EliotJones, line 19: Could these property setters be internal to stop consumers messing with them?

EliotJones:
@BobLd on `IResourceStore` an parsing options afaict from a quick scan it should be fine to move them to the constructor of the resource store instance since we have the options at that stage

src/UglyToad.PdfPig/Content/PageTreeMembers.cs
BobLd: well spotted, now updated

BobLd:
@EliotJones good point on the resource store instance, I'll look into it and create another PR for that. Let's keep this one simple

EliotJones (approved):
LGTM

PR #719, opened 2023-10-23
Add IPageFactory to the public API, remove InternalParsingOptions

BobLd:
- Add `IPageFactory` to the public API
- Remove `InternalParsingOptions`. I went for this option but very happy to have your input. If you want to keep a read-only version on the parsing options, we could go for a public `ImmutableParsingOptions` (replacing `InternalParsingOptions`) with only private setter. Also, I have not added the `SkipMissingXObjects` property to the parsing option class as it was never used and set to the `SkipMissingFonts` value.


EliotJones:
Just to check, since the (Internal)ParsingOptions moves from the interface to the (internal) class constructor anyway is it necessary to actually remove it? Or is this somewhere the type is causing enough pain in enough places that you thought it's better to remove it entirely anyway?

BobLd:
@EliotJones the problem with the options remains, because even if we pass the option in the constructors, the new  `AddPageFactory<TPage, TPageFactory>()` (to come) method will expect a constructor with the same signature as the default `PageFactory` (i.e. with the parsing options).

Below is my POC PR, the options are not in the constructor, but are in the `Create()`:
https://github.com/UglyToad/PdfPig/blob/843ea8fb00f4f1bdef27c9b92579d686237bfcb8/src/UglyToad.PdfPig/Content/Pages.cs#L109C4-L119C10

I believe the new (external/users) implementations of `IPageFactory` should also have access to these options (keeping in mind the logger is in the options, even if in the current code base it is passed to the `PageFactory` constructor). But if you don't think so (or think we can do that later), I can leave the `InternalParsingOptions` in the (internal) `PageFactory` constructor, and the (external/users) implemetations of the `IPageFactory` will have a constructor that does not have any options, and won't be able to access them:

In `Pages.cs`, with this PR we would have something like:
```csharp
internal void AddPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
{
          var defaultPageFactory = (PageFactory)pageFactoryCache[typeof(Page)];

          var pageFactory = (IPageFactory<TPage>)Activator.CreateInstance(typeof(TPageFactory),
                  defaultPageFactory.PdfScanner,
                  defaultPageFactory.ResourceStore,
                  defaultPageFactory.FilterProvider,
                  defaultPageFactory.PageContentParser,
                  defaultPageFactory.ParsingOptions);
          AddPageFactory(pageFactory);
}
```

but we could ignore the options (I would not recommand that), and have:

```csharp
internal void AddPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
{
          var defaultPageFactory = (PageFactory)pageFactoryCache[typeof(Page)];

          var pageFactory = (IPageFactory<TPage>)Activator.CreateInstance(typeof(TPageFactory),
                  defaultPageFactory.PdfScanner,
                  defaultPageFactory.ResourceStore,
                  defaultPageFactory.FilterProvider,
                  defaultPageFactory.PageContentParser);
          AddPageFactory(pageFactory);
}
```

Maybe another option is to still remove the `InternalParsingOptions`, but add a constructor to the `ParsingOptions` class that will take the properties as input and make the property setters private. Doing so we ensure the options are read only. I will just need to make sure the passwords array is properly set

EliotJones (approved):
Fine by me, to be honest if people mess around with the options mid parsing that's their issue :D

PR #742, opened 2023-12-15
Fix Type 0 font to use CID font's correct matrix and Charset when Encoding is null

BobLd:
- Fix Type 0 font to use CID font's correct matrix and `Charset` when `Encoding` is null
- Minor optimisation in fonts `TryGetNormalisedPath()`


EliotJones (approved):
I don't quite understand the changes here but they look more in line with the spec.

BobLd:
@EliotJones I have to admit it's the same for me, the pdf documentation is far from being clear. It's still a draft PR and I need way more tests and I also need to add unit tests. I want to make sure the PR does not introduce regression.

BobLd:
Replaced by #757

Issue #664, opened 2023-07-24
Linker warnings from the AES code when publishing a NativeAOT application


Numpsy:
Hi,

I've been doing some tests with publishing some apps as NativeAOT to see what works and when I built an app that references PdfPig I noticed this in the build logs:

```
UglyToad.PdfPig.Encryption.EncryptionHandler.CalculateKeyRevisions5And6(Byte[],EncryptionDictionary,Boolean): Using member 'System.Security.Cryptography.Aes.Create(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The default algorithm implementations might be removed, use strong type references like 'RSA.Create()' instead.
```

And then I also notice that the MS docs at https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aes.create?view=net-7.0#system-security-cryptography-aes-create(system-string) say that that overload of ```Create``` has been obsoleted in .NET 7.
So I just wondered - is there a need to use that version of Create rather than the base one?

BobLd:
I see other warning with `ReflectionGraphicsStateOperationFactory` and it actually prevent my NativeAOT to run. I'll give it a go with a PR soon

Numpsy:
@BobLd The title issue with AES is hopefully fixed by #747, but do you want the issue left open for the ```ReflectionGraphicsStateOperationFactory``` one?

BobLd:
@Numpsy I think we can close this one, as it was specific to aes. Thanks a lot for your help with that. I'll try to look into the other issues for Native AOT but any help would be appreciated 😁

Numpsy:
I've done a bit in that area on other projects by adding attributes and such so I can have a look at some point, though I don't know when I'll have time.

For the record, you get warnings about the graphics code, and a similar one in AddPageFactory, as regular build warnings if you enable the ```EnableTrimAnalyzer``` property in the .net6.0 build, so maybe that's something to consider doing?

PR #750, opened 2024-01-10
Add a DynamicallyAccessedMembers attribute to AddPageFactory<TPage, T…

Numpsy:
…PageFactory>

A bit of a follow up to #664 - 

If I build the core library with the .NET 6.0 trimming analyzer enabled, then I get this warning from the PageFactory code:
```
PdfPig\src\UglyToad.PdfPig\Content\Pages.cs(122,31,129,15): warning IL2090: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Type.GetConstructor(Type[])'. The generic parameter 'TPageFactory' of 'UglyToad.PdfPig.Content.Pages.AddPageFactory<TPage, TPageFactory>()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```
which is warning that dynamic access to the constructors could fail if the constructors have got trimmed out.
This is an attempt to fix that by adding a ```System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers``` attribute to ```TPageFactory```.

Note: I fully qualified the names because when I tried ```using System.Diagnostics.CodeAnalysis``` I got conflicts between the ```NotNull``` attribute in there and the Jetbrains one already in use. (maybe that can be handled in a different way)

### DIFF
--- a/src/UglyToad.PdfPig/Content/Pages.cs
+++ b/src/UglyToad.PdfPig/Content/Pages.cs
@@ -117,7 +117,11 @@
             pageFactoryCache.Add(type, pageFactory);
         }
 
+#if NET6_0_OR_GREATER
+        internal void AddPageFactory<TPage, [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] TPageFactory>() where TPageFactory : IPageFactory<TPage>
+#else
         internal void AddPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
+#endif
         {
             var constructor = typeof(TPageFactory).GetConstructor(new[]
             {
--- a/src/UglyToad.PdfPig/PdfDocument.cs
+++ b/src/UglyToad.PdfPig/PdfDocument.cs
@@ -148,7 +148,11 @@
         /// </summary>
         /// <typeparam name="TPage"></typeparam>
         /// <typeparam name="TPageFactory"></typeparam>
+#if NET6_0_OR_GREATER
+        public void AddPageFactory<TPage, [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] TPageFactory>() where TPageFactory : IPageFactory<TPage>
+#else
         public void AddPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
+#endif
         {
             pages.AddPageFactory<TPage, TPageFactory>();
         }


BobLd:
@Numpsy thanks a lot for that! I faced the same issue with AOT and I confirm I had also to fully qualify names too.

Feel free to remove the draft status and I'll approve and merge asap.

 Thanks again.

BobLd (approved):

EliotJones:
@BobLd I wonder if this use of reflection could be swapped to something like:

```
public delegate TPageFactory CreatePageFactory(PdfScanner scanner, ResourceStore store, FilterProvider...);

internal void AddPageFactory<TPage, TPageFactory>(CreatePageFactory creator) where TPageFactory : IPageFactory<TPage>
{
    var instance = creator(defaultPageFactory.PdfScanner, ...);
    AddPageFactory(instance);
}
```

So it takes in a delegate/Func that takes the constructor parameters and returns an instance of a page factory, that way reflection can be avoided entirely I think?

Numpsy:
I think that using factory functions instead of reflection is one of the ways of avoiding issues with trimming and AOT, could possibly have less overhead as well (if that'd be noticable in this case)

BobLd:
@EliotJones I think you are right, but my personal taste here is to use reflection for its convenience in this particular case. And just to confirm, adding attributes indeed allows proper trimming (at least in net8) and allows full AOT code to run.

In my case I have done as follow (not sure everything is necessary though as it came from the Microsoft How-to)
```csharp
#if NET8_0_OR_GREATER
        internal void AddPageFactory<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.PublicMethods)] TPage, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.PublicMethods)] TPageFactory>() where TPageFactory : IPageFactory<TPage>
#else
        internal void AddPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
#endif
```

Let me know what you think but I'm happy to merge as is if that's fine with you

Numpsy:
I was using NET6_0_OR_GREATER rather than NET8_0_OR_GREATER because, although AOT requires .NET 8, single file publishing with trimming works on .NET 6/7 and the attributes are useful there as well.

Not sure about the attributes on TPage as well as TPageFactory - the analyzer warnings are just about TPageFactory as that's what's being instantiated by reflection, but I haven't done detailed testing to see if there are actually further issues beyone that (I don't think the analyzers find all possible issues just when compiling the library itself)

BobLd:
@Numpsy yes using NET6_0_OR_GREATER instead of NET8_0_OR_GREATER is much better, I was just testing with net8 on my side. 

Regarding the attributes on TPage, let's try like that for the moment and see how it behaves. I'm fine merging as is if @EliotJones is gtg

EliotJones:
I'm all good here, thanks for the work

PR #757, opened 2024-01-15
Fix font glyph path

BobLd:
This replaces #742

Sorry for the long PR. Many file modified are just tidy up.

Regarding how font matrices are actually supposed to work, this is not clear as per the documenation. have a look here if you want to see some discussion:
- https://github.com/veraPDF/veraPDF-library/issues/1010
- https://bugs.ghostscript.com/show_bug.cgi?id=690724

All my changes are backed by visual tests (for both glyph paths and bounding boxes)

Added test documents either come from:
- https://github.com/veraPDF/veraPDF-library/issues/1010
- https://pdfa.org/stressful-pdf-corpus/


src/UglyToad.PdfPig.Fonts/CompactFontFormat/CharStrings/Type2CharStrings.cs
```diff
@@ -11,7 +11,7 @@
     /// as the local (per font) and global (per font set) subroutines.
     /// The CharStrings are lazily evaluated.
     /// </summary>
-    internal class Type2CharStrings
+    internal sealed class Type2CharStrings
```
EliotJones, line 14: Any particular reason to seal these? Does it provide performance benefits?

src/UglyToad.PdfPig.Fonts/CompactFontFormat/Charsets/CompactFontFormatCharset.cs
```diff
@@ -37,7 +37,7 @@ public virtual string GetNameByGlyphId(int glyphId)
 
         public virtual string GetNameByStringId(int stringId)
         {
-            return GlyphIdToStringIdAndName.Single(x => x.Value.stringId == stringId).Value.name;
+            return GlyphIdToStringIdAndName.SingleOrDefault(x => x.Value.stringId == stringId).Value.name;
```
EliotJones, line 40: Would this not throw where we access `.Value.name` or is the thing being accessed a tuple or struct anyway?

src/UglyToad.PdfPig.Fonts/CompactFontFormat/CharStrings/Type2CharStrings.cs
BobLd: It can offer performance improvements, see here https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter

As a general rule of thumb, I always seal my classes, with the possibility to unseal it if need be. My reasoning is "_I need a reason for a class to not be sealed_", instead of "_I need a reason for a class to be sealed_".

But let me know what you think, happy to discuss

EliotJones: Thanks, in this case it's fine since these are internal classes. I think for public facing classes it would be a breaking API change to start sealing them if people are relying on creating derived types which we wouldn't be able to detect. I'll flag any public instances.

BobLd:
@EliotJones I've pushed a commit to fix a compilation warning.

Regarding your point about sealed classes, the only 2 public ones I sealed I could find are `CompactFontFormatFontCollection` and `GlyphList`, let me know if you want me to unseal them (I'm not sure how could people derive and use classes based on those, but I really could be wrong)

src/UglyToad.PdfPig.Fonts/CompactFontFormat/CompactFontFormatFontCollection.cs
```diff
@@ -8,7 +8,7 @@
     /// A Compact Font Format (CFF) font program as described in The Compact Font Format specification (Adobe Technical Note #5176).
     /// A CFF font may contain multiple fonts and achieves compression by sharing details between fonts in the set.
     /// </summary>
-    public class CompactFontFormatFontCollection
+    public sealed class CompactFontFormatFontCollection
```
EliotJones, line 11: this will change the public API

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
```diff
@@ -1239,6 +1239,12 @@ public override IColor GetColor(params double[] values)
 
             // TODO - use ICC profile
 
+            for (int c = 0; c < values.Length; c++)
+            {
+                int i = 2 * c;
+                values[c] = PdfFunction.ClipToRange(values[c], Range[i], Range[i + 1]);
```
EliotJones, line 1245: Out of interest what is happening here?

EliotJones (commented):
just a couple of places becoming sealed that would change the public API, otherwise LGTM even though I'm still not 100% on what's happening but if the tests look good then it should be right

src/UglyToad.PdfPig.Fonts/CompactFontFormat/CompactFontFormatFontCollection.cs
BobLd: unsealed

src/UglyToad.PdfPig.Fonts/GlyphList.cs
```diff
@@ -9,9 +9,12 @@
     /// <summary>
     /// A list which maps PostScript glyph names to unicode values.
     /// </summary>
-    public class GlyphList
+    public sealed class GlyphList
```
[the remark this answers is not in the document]
BobLd: unsealed

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
BobLd: This is tested in `RepresentativeWindEnergyDeals()` in the `GenerateLetterGlyphImages` (doument is GHOSTSCRIPT-698168-0.pdf)

In this particular test, the color space is ICC, but we don't support it. So we use the Alternate CC.


> Range
> (Optional) An array of 2 × N numbers [𝑚𝑖𝑛0 𝑚𝑎𝑥0 𝑚𝑖𝑛1 𝑚𝑎𝑥1…] that shall specify the minimum and maximum valid values of the corresponding colour components. These values shall match the information in the ICC profile. Default value: [0.0 1.0 0.0 1.0…].

BobLd:
I agree with you, this is not exactly clear what happens. And reading the discussions in the links I've put, it seems we are not the only ones. I tried to document the code to reflect that

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
BobLd: I think this is how they do in PdfBox, where they clamp values too.

https://github.com/apache/pdfbox/blob/33ebd39096a650df157a846d9e8c631a1e3631ab/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/graphics/color/PDICCBased.java#L291C13-L291C75

Again, happy for you to tell me if you understand that differently

EliotJones (approved):

Issue #758, opened 2024-01-16
PdfPig warning are wasting trace


marcRobertHabib:
Thanks for your work,

We are writting an application reading thousands of PDF from an external source, and the application uses System.Diagnostics.Trace to write a log file. 
I don't know why but we are getting for each document a lot of warnings like below in the trace:
Warning: 0 : 16/01/2024 14:54:47: Table with invalid checksum found in TrueType font file.
    16/01/2024 14:54:47: Callstack=   at System.Environment.get_StackTrace()
   at System.Diagnostics.TraceListener.WriteFooter(TraceEventCache eventCache)
   at System.Diagnostics.TraceInternal.TraceEvent(TraceEventType eventType, Int32 id, String format, Object[] args)
   at UglyToad.PdfPig.Fonts.TrueType.Parser.TableParser.Parse[T](TrueTypeHeaderTable table, TrueTypeDataBytes data, Builder register)
   at UglyToad.PdfPig.Fonts.TrueType.Parser.TrueTypeFontParser.ParseTables(Single version, IReadOnlyDictionary`2 tables, TrueTypeDataBytes data)
   at UglyToad.PdfPig.Fonts.TrueType.Parser.TrueTypeFontParser.Parse(TrueTypeDataBytes data)
   at UglyToad.PdfPig.PdfFonts.Parser.Parts.CidFontFactory.ReadDescriptorFile(FontDescriptor descriptor)
   at UglyToad.PdfPig.PdfFonts.Parser.Parts.CidFontFactory.Generate(DictionaryToken dictionary)
   at UglyToad.PdfPig.PdfFonts.Parser.Handlers.Type0FontHandler.ParseDescendant(DictionaryToken dictionary)
   at UglyToad.PdfPig.PdfFonts.Parser.Handlers.Type0FontHandler...

And the log file becomes huge. Is there a way to prevent these messages?

Thanks

BobLd:
@marcRobertHabib Are you using the NuGet package in your app? Or are you referencing the code directly? Are you getting these trace when running in Debug mode or Release mode?

marcRobertHabib:
I am using Nuget and PdfPig 0.1.7

I am getting these messages in both release and debug mode.

Thanks

מרק חביב

BobLd:
@EliotJones Unless there's an actual reason why these logs are Trace and not Debug in PdfPig, I'll create a PR to change that.

As to why there's an exception here, I'm not sure.

[...]

PR #760, opened 2024-01-16
Replace Trade by Debug

BobLd:
See #758

### DIFF
--- a/src/UglyToad.PdfPig.Fonts/TrueType/Parser/TableParser.cs
+++ b/src/UglyToad.PdfPig.Fonts/TrueType/Parser/TableParser.cs
@@ -14,7 +14,7 @@
 
         public static T Parse<T>(TrueTypeHeaderTable table, TrueTypeDataBytes data, TableRegister.Builder register) where T : ITrueTypeTable
         {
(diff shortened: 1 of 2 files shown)


src/UglyToad.PdfPig.Fonts/TrueType/Parser/TableParser.cs
```diff
@@ -25,7 +25,7 @@ internal static class TableParser
 
             if (sum != table.CheckSum)
             {
-                Trace.TraceWarning("Table with invalid checksum found in TrueType font file.");
+                Debug.Write("Table with invalid checksum found in TrueType font file.");
```
EliotJones, line 28: Cool, didn't know you could just do `Write`, I thought it was asserts only, thanks

EliotJones (approved):

src/UglyToad.PdfPig.Fonts/TrueType/Parser/TableParser.cs
BobLd: I meant to use `WriteLine` and not `Write`. I've changed that now

PR #762, opened 2024-01-21
Use double for pdf version instead of decimal

BobLd:


EliotJones (commented):
TBH I'm not sure double makes much sense here, since you don't need to do math on the version ever really it seems to make sense to keep it as decimal, since it is always exactly 1.1, 1.2, 1.5, etc. What do you think?

BobLd:
I agree with you that there's no maths done here but I still think it's good to move away from `decimal` in this case, and `double` gives enough precision.

Also, PdfBox uses `float` for the pdf version, see here https://github.com/apache/pdfbox/blob/d8d401be7d261422f1590869ceabc82499784dcd/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/PDDocument.java#L1382

EliotJones:
I'm still conflicted on this one, if the field was only used internally I'd say the benefit of being able to check `decimal` -> `double` migration progress easily outweighs the cost. But since version is part of the public API I'm not sure there's a compelling reason to change it for this field specifically. I'd imagine Java's lack of a `decimal` equivalent (other than `BigDecimal` is why PDFBox uses `float`. I'm not sure, let me think on it some more.

EliotJones:
Sorry I've been away on vacation. Since other changes will move to `double` anyway and for the sake of consistency I'm fine with this

EliotJones (approved):

Issue #771, opened 2024-02-06
Security issues


DanMcNamee42:
I was investigating using this package for use on a work project, and part of our due diligence before using packages is to run the package through a scan for vulnerabilities.  Our scan found multiple security vulnerabilities with the old versions of .NetCore and Newtonsoft.Json that should be fixed by updating the packages to the current versions.

I like the package, it does exactly what we need, but with the vulnerabilities, I won't be able to get it approved.  Any chance of getting the packages you are using updated to address the vulnerabilities?

Thanks!

Numpsy:
When you say 'the package', so you mean the PdfPig nuget package or something else in the repo? (the main lib seems to only depend on System.ValueTuple, and then only on .NET 4.x, though adding a .NET 4.7.1 target might remove that for 4.7.1+ consumers)

Numpsy:
There are references to .NET Standard.Library 1.6.1 from the tests though:
![image](https://github.com/UglyToad/PdfPig/assets/1178570/3523c3da-896d-4d65-9071-47d786386db3)

which does cause complaints from [certain tooling](https://github.com/RicoSuter/NSwag/pull/4716#issuecomment-1928068824).

DanMcNamee42:
Sorry I wasn't clear.  It's with NuGet packages that are included in PdfPig.  They could be packages that are pulling in other dependencies that are out of date.  Looking at the NuGet packages that are in PdfPig now, all of them are out of date, with CodeCov being deprecated and no longer maintained.  Possibly updating the NuGet packages will update the dependencies..

Numpsy:
there's already a discussion about the coverage thing in #755

DanMcNamee42:
Thank you!  Hopefully there will be a discussion about updating the other packages as well.

Numpsy:
The dependencies in the test project have been updated and minimzed now, which has removed transitive references to a load of old stuff, though the test project itself still runs against .NET Core 2.1 as well as 6.0

DanMcNamee42:
Thank you!

DanMcNamee42:
It all looks great now.  Thank you!

PR #772, opened 2024-02-06
Retarget the unit test project from .NET Core 2.0 to 2.1, and update the test SDK

Numpsy:
refs #771

I'm not entirely certain that this will work even though version 17.3.3 still lists older versions of .NET Core as supported TFMs (anything newer requires a newer .NET target in the test project, which is presently still targetting .NET Core 2,0 as well as 6.0)

Anyway - the idea is that it takes out references to a great deal of old libraries - 

with version 16.2.0 of the test SDK:

image

and with 17.3.3:

image

including an older version of Newtonsoft.Json which is listed as having security issues (i can't say whether it actually does in this usage)

### DIFF
--- a/src/UglyToad.PdfPig.Tests/UglyToad.PdfPig.Tests.csproj
+++ b/src/UglyToad.PdfPig.Tests/UglyToad.PdfPig.Tests.csproj
@@ -88,7 +88,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.3" />
     <PackageReference Include="SkiaSharp" Version="2.88.7" />
     <PackageReference Include="xunit" Version="2.6.6" />
     <PackageReference Include="xunit.extensibility.execution" Version="2.6.6" />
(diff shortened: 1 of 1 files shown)


Numpsy:
Well that's a useful error
```
Failed to load the dll from [], HRESULT: 0x80070057
```
:-(

BobLd:
@Numpsy Seems like updating the test project from `<TargetFrameworks>netcoreapp2.0;net6.0</TargetFrameworks>` to `<TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>` fixes the issue. From what I see it will still test against the `netstandard2.0` version of the library, so we are fine.

Also seems that the Microsoft.NET.Test.Sdk package can then be updated to it's latest version (17.9.0)

Numpsy:
I tried having a go locally updating it to Core 2.1 and that seemed ok as well, but I'll do a PR to change it to 3.1 if you're ok with that.

BobLd:
okay great! whatever you prefer really

Numpsy:
Actually, It looke like the CI build pipeline is already installing the 2.1 SDK rather than the 2.0 one - https://github.com/UglyToad/PdfPig/blob/47cfcdb690da198482fce24fee6f4358e5555cb1/.github/workflows/build_and_test.yml#L19 so maybe getting the test project matched up is the place to start?

BobLd:
@Numpsy yes sounds good to me

Numpsy:
Ok, that looks to have worked, and I think that's the newest versions of everything that can be used without going up to .NET Core 3.1

BobLd (approved):

BobLd:
Amazing @Numpsy, thanks a lot!

PR #776, opened 2024-02-07
Fix code analyzer suggestions coming from the updated xUnit packages

Numpsy:
It seems that the update to xUnit has resulted in a number of analyzer suggestions that are coming out as warnings in the CI build:

image

So I'm having a go at making the suggested changes to fix the warnings


BobLd (approved):

BobLd:
Thanks a lot @Numpsy, that's great!

PR #787, opened 2024-03-03
Use double in fonts instead of decimals and tidy up remaining decimals

BobLd:
Final PR to replace the use of `decimal`s with `double`s


EliotJones (approved):
Congratulations on vanquishing your foe :D

PR #799, opened 2024-03-15
Improve Code Quality 2

iamcarbon:
This PR:

- Spanifies internal Adler32Checksum and Crc32 methods
- Use switch expressions
- Enables nullable annotations (this just allows us to begin annotating, before turning on fully)
- Use collection expressions in more places where semantics don't change
- Prefer is null to == null (this ensures that we compare directly against null, and don't call an overloaded == operator)


iamcarbon:
@EliotJones @BobLd Ready for feedback / review.

BobLd:
@iamcarbon 2 comments, only one needs an update

BobLd (approved):

PR #805, opened 2024-03-17
Improve Code Quality 3

iamcarbon:
This PR:

- Throws when inverting a matrix with a determinate of 0 instead of returning null. Our only caller expects a non-null result and will currently throw a NRE (NullReferenceException). This error should be more informative.
- Used vectorized hex encoding on modern .NET
- Replaces various Substrings with spans to eliminate various alllocations
- Utilized the vectored Span.Fill method and eliminates a fill helper


iamcarbon:
@ BobLd Ready for feedback / review.

src/UglyToad.PdfPig/Util/Matrix3x3.cs
```diff
@@ -69,14 +69,14 @@ public IEnumerator<double> GetEnumerator()
...
-        public Matrix3x3? Inverse()
+        public Matrix3x3 Inverse()
         {
             var determinant = GetDeterminant();
 
             // No inverse matrix exists when determinant is zero
             if (determinant == 0)
             {
-                return null;
+                throw new InvalidOperationException("May not inverse a matrix with a determinant of 0.");
```
iamcarbon, line 79: NOTE: our only caller will throw an NRE if this method returns null. Throwing early should give us a better debugging experience.

BobLd (approved):

PR #815, opened 2024-04-01
Spanify work 2

iamcarbon:
This PR:

- Update the filters and their callers to utilize ReadOnly{Memory,Span}<byte>
- Polyfills Stream.Write(ReadOnlySpan<byte>)
- Updates various methods to accept ReadOnlySpan<byte>
- Eliminates various allocations
- Makes QuadPointsQuadrilateral a readonly struct
- Spanifies PngBuilder
- Spanifies InputBytes

### DIFF
--- a/src/UglyToad.PdfPig.Tokenization/Scanner/CoreTokenScanner.cs
+++ b/src/UglyToad.PdfPig.Tokenization/Scanner/CoreTokenScanner.cs
@@ -210,7 +210,7 @@
                         // Special case handling for inline images.
                         var imageData = ReadInlineImageData();
                         isInInlineImage = false;
-                        CurrentToken = new InlineImageDataToken(imageData);
+                        CurrentToken = new InlineImageDataToken(new ReadOnlyMemory<byte>([..imageData]));
                         hasBytePreRead = false;
                         return true;
                     }
@@ -287,7 +287,7 @@
             return data;
         }
 
-        private IReadOnlyList<byte> ReadInlineImageData()
+        private List<byte> ReadInlineImageData()
         {
             // The ID operator should be followed by a single white-space character, and the next character is interpreted
             // as the first byte of image data. 
@@ -305,7 +305,6 @@
         {
             const byte lastPlainText = 127;
             const byte space = 32;
-
 
             var imageData = new List<byte>();
             byte prevByte = 0;
--- a/src/UglyToad.PdfPig/Content/BasePageFactory.cs
+++ b/src/UglyToad.PdfPig/Content/BasePageFactory.cs
@@ -118,7 +118,7 @@
             }
             else if (DirectObjectFinder.TryGet<ArrayToken>(contents, PdfScanner, out var array))
             {
-                using var bytes = new ArrayPoolBufferWriter<byte>();
+                using var bytes = new ArrayPoolBufferWriter<byte>(1024 * 64);
 
                 for (var i = 0; i < array.Data.Count; i++)
                 {
@@ -144,7 +144,7 @@
                     }
                 }
 
-                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes.WrittenSpan.ToArray());
+                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes.WrittenMemory);
             }
             else
             {
--- a/src/UglyToad.PdfPig/Content/BasePageFactory.cs
+++ b/src/UglyToad.PdfPig/Content/BasePageFactory.cs
@@ -118,7 +118,7 @@
             }
             else if (DirectObjectFinder.TryGet<ArrayToken>(contents, PdfScanner, out var array))
             {
-                var bytes = new List<byte>();
+                using var bytes = new ArrayPoolBufferWriter<byte>();
 
                 for (var i = 0; i < array.Data.Count; i++)
                 {
@@ -136,15 +136,15 @@
                         throw new InvalidOperationException($"Could not find the contents for object {obj}.");
                     }
 
-                    bytes.AddRange(contentStream.Decode(FilterProvider, PdfScanner));
+                    bytes.Write(contentStream.Decode(FilterProvider, PdfScanner).Span);
 
                     if (i < array.Data.Count - 1)
                     {
-                        bytes.Add((byte)'\n');
+                        bytes.Write((byte)'\n');
                     }
                 }
 
-                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes);
+                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes.WrittenSpan.ToArray());
             }
             else
             {
@@ -177,18 +177,18 @@
             UserSpaceUnit userSpaceUnit,
             PageRotationDegrees rotation,
             TransformationMatrix initialMatrix,
-            IReadOnlyList<byte>? contentBytes)
+            ReadOnlyMemory<byte> contentBytes)
         {
             IReadOnlyList<IGraphicsStateOperation> operations;
 
-            if (contentBytes is null || contentBytes.Count == 0)
+            if (contentBytes.IsEmpty)
             {
                 operations = Array.Empty<IGraphicsStateOperation>();
             }
             else
             {
                 operations = PageContentParser.Parse(pageNumber,
-                    new ByteArrayInputBytes(contentBytes),
+                    new MemoryInputBytes(contentBytes),
                     ParsingOptions.Logger);
             }
 
--- a/src/UglyToad.PdfPig/Content/IPdfImage.cs
+++ b/src/UglyToad.PdfPig/Content/IPdfImage.cs
@@ -40,6 +40,11 @@
         ReadOnlyMemory<byte> RawMemory { get; }
 
         /// <summary>
+        /// The encoded memory span of the image with all filters still applied.
+        /// </summary>
+        ReadOnlySpan<byte> RawBytes { get; }
+
+        /// <summary>
         /// The color rendering intent to be used when rendering the image.
         /// </summary>
         RenderingIntent RenderingIntent { get; }
@@ -90,10 +95,10 @@
         ColorSpaceDetails? ColorSpaceDetails { get; }
 
         /// <summary>
-        /// Get the decoded bytes of the image if applicable. For JPEG images and some other types the
+        /// Get the decoded memory of the image if applicable. For JPEG images and some other types the
         /// <see cref="RawMemory"/> should be used directly.
         /// </summary>
-        bool TryGetMemory(out ReadOnlyMemory<byte> bytes);
+        bool TryGetMemory(out ReadOnlyMemory<byte> memory);
 
         /// <summary>
         /// Try to convert the image to PNG. Doesn't support conversion of JPG to PNG.
--- a/src/UglyToad.PdfPig/Content/IPdfImage.cs
+++ b/src/UglyToad.PdfPig/Content/IPdfImage.cs
@@ -98,7 +98,7 @@
         /// Get the decoded memory of the image if applicable. For JPEG images and some other types the
         /// <see cref="RawMemory"/> should be used directly.
         /// </summary>
-        bool TryGetMemory(out ReadOnlyMemory<byte> memory);
+        bool TryGetBytesAsMemory(out ReadOnlyMemory<byte> memory);
 
         /// <summary>
         /// Try to convert the image to PNG. Doesn't support conversion of JPG to PNG.
--- a/src/UglyToad.PdfPig/Content/IPdfImage.cs
+++ b/src/UglyToad.PdfPig/Content/IPdfImage.cs
@@ -1,5 +1,6 @@
 ﻿namespace UglyToad.PdfPig.Content
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using Core;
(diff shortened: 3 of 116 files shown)


iamcarbon:
@BobLd Ready for review. This is the bulk of the spanification work and sets us up nicely for further optimization work.

src/UglyToad.PdfPig.Tokenization/Scanner/CoreTokenScanner.cs
```diff
@@ -287,7 +287,7 @@ public IReadOnlyList<byte> RecoverFromIncorrectEndImage(long lastEndImageOffset)
             return data;
         }
 
-        private IReadOnlyList<byte> ReadInlineImageData()
+        private List<byte> ReadInlineImageData()
```
BobLd, line 290: why use `List<byte>`? Is it necessary?

src/UglyToad.PdfPig/Content/BasePageFactory.cs
```diff
@@ -136,15 +136,15 @@ public abstract class BasePageFactory<TPage> : IPageFactory<TPage>
...
 
                     if (i < array.Data.Count - 1)
                     {
-                        bytes.Add((byte)'\n');
+                        bytes.Write((byte)'\n');
                     }
                 }
 
-                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes);
+                page = ProcessPageInternal(number, dictionary, namedDestinations, mediaBox, cropBox, userSpaceUnit, rotation, initialMatrix, bytes.WrittenSpan.ToArray());
```
BobLd, line 147: I assume you will remove the `ToArray()` in later commit

src/UglyToad.PdfPig/Content/InlineImage.cs
```diff
@@ -50,7 +49,7 @@ public class InlineImage : IPdfImage
         public bool Interpolate { get; }
 
         /// <inheritdoc />
-        public IReadOnlyList<byte> RawBytes { get; }
+        public ReadOnlyMemory<byte> RawMemory { get; }
```
BobLd, line 52: Use `RawBytes` - see other comment. Unless you intend to expose `public ReadOnlySpan<byte> RawBytes => RawMemory.Span`, if yes - then let's put both

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
```diff
@@ -954,15 +983,17 @@ private RGBColor TransformToRGB((double A, double B, double C) colorAbc)
         }
 
         /// <inheritdoc/>
-        internal override IReadOnlyList<byte> Transform(IReadOnlyList<byte> decoded)
+        internal override ReadOnlySpan<byte> Transform(ReadOnlySpan<byte> decoded)
         {
-            var transformed = new List<byte>();
-            for (var i = 0; i < decoded.Count; i += 3)
+            var transformed = new byte[decoded.Length];
```
BobLd, line 988: Note to self - Check if that always work

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
```diff
@@ -1074,15 +1105,17 @@ private RGBColor TransformToRGB((double A, double B, double C) colorAbc)
...
```
BobLd, line 1110: Note to self - same here

src/UglyToad.PdfPig/Images/Png/ImageHeader.cs
```diff
@@ -8,9 +8,9 @@
     /// </summary>
     internal readonly struct ImageHeader
     {
-        internal static readonly byte[] HeaderBytes = {
+        internal static ReadOnlySpan<byte> HeaderBytes => [
```
BobLd, line 11: should be better to use 
```
internal static readonly ReadOnlySpan<byte> HeaderBytes = [ 73, 72, 68, 82 ];
```

src/UglyToad.PdfPig/Util/ColorSpaceDetailsParser.cs
```diff
@@ -56,7 +55,7 @@ internal static class ColorSpaceDetailsParser
                 var colorSpaceDetails = GetColorSpaceDetails(colorSpace, imageDictionary.Without(NameToken.Filter).Without(NameToken.F), scanner, resourceStore, filterProvider, true);
 
                 var decodeRaw = imageDictionary.GetObjectOrDefault(NameToken.Decode, NameToken.D) as ArrayToken ?? new ArrayToken([]);
-                var decode = decodeRaw.Data.OfType<NumericToken>().Select(x => x.Double).ToArray();
+                var decode = decodeRaw.Data.OfType<NumericToken>().Select(static x => x.Double).ToArray();
```
BobLd, line 58: what does `static` means in this context?

BobLd (changes requested):

BobLd:
@iamcarbon first round of review - thanks again for the hard work

src/UglyToad.PdfPig/Util/ColorSpaceDetailsParser.cs
iamcarbon: This ensures that the inner method doesn't capture (or allocate a closure). It also allows you to reuse variable names inside the static scope. 

The code gen in this case is the same, but these are nice to make sure we don't accidentally capture inside the lambda.

src/UglyToad.PdfPig/Images/Png/ImageHeader.cs
iamcarbon: Updated!

src/UglyToad.PdfPig/Content/InlineImage.cs
iamcarbon: Doubled up RawBytes, which should minimize any breaking changes after a compilation.

src/UglyToad.PdfPig/Content/BasePageFactory.cs
iamcarbon: Eliminated! This was potentially a LOH allocation, so great to kill.

src/UglyToad.PdfPig.Tokenization/Scanner/CoreTokenScanner.cs
iamcarbon: This eliminates possible virtual call and wasn't on the public API, but agree it breaks consistency with the rest of the codebase. Let me know if we should revert this one.

src/UglyToad.PdfPig/Content/IPdfImage.cs
```diff
@@ -34,9 +35,9 @@ public interface IPdfImage
         int BitsPerComponent { get; }
 
         /// <summary>
-        /// The encoded bytes of the image with all filters still applied.
+        /// The encoded memory of the image with all filters still applied.
         /// </summary>
-        IReadOnlyList<byte> RawBytes { get; }
+        ReadOnlyMemory<byte> RawMemory { get; }
```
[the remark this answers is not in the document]
iamcarbon: Doubled up RawBytes here.

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
iamcarbon: should have the identical behavior as the List, without any additional copying as the list grows.

iamcarbon:
@BobLd Thanks for the review. Let me know if you spot anything else when you look over things in depth. I did another pass over everything, and is looking good from my side.

BobLd:
@iamcarbon I ran some benchmarks again following my previous comment. It seems the (alloc memory) performance decrease comes from your latest commit ([Remove unused StreamHelper methods](https://github.com/UglyToad/PdfPig/pull/815/commits/8314f411764d81ceb9f282214f63645b4b13b4c9))

memory2 (latest)
image

memory2 (latest commit is `Eliminate allocation in BasePageFactory`)
image

src/UglyToad.PdfPig/Util/PatternParser.cs
```diff
@@ -20,7 +20,7 @@ public static PatternColor Create(IToken pattern, IPdfTokenScanner scanner, IRes
             if (DirectObjectFinder.TryGet(pattern, scanner, out StreamToken? fs))
             {
                 patternDictionary = fs.StreamDictionary;
-                patternStream = new StreamToken(fs.StreamDictionary, fs.Decode(filterProvider, scanner));
+                patternStream = new StreamToken(fs.StreamDictionary, fs.Decode(filterProvider, scanner).ToArray());
```
BobLd, line 23: I don't think you need thew `ToArray()¬, you can use the `ReadOnlyMemory` directly

BobLd: (same for all `fs.Decode(filterProvider, scanner)`)

src/UglyToad.PdfPig/Content/IPdfImage.cs
```diff
@@ -89,10 +95,10 @@ public interface IPdfImage
         ColorSpaceDetails? ColorSpaceDetails { get; }
 
         /// <summary>
-        /// Get the decoded bytes of the image if applicable. For JPEG images and some other types the
-        /// <see cref="RawBytes"/> should be used directly.
+        /// Get the decoded memory of the image if applicable. For JPEG images and some other types the
+        /// <see cref="RawMemory"/> should be used directly.
         /// </summary>
-        bool TryGetBytes([NotNullWhen(true)] out IReadOnlyList<byte>? bytes);
+        bool TryGetMemory(out ReadOnlyMemory<byte> memory);
```
BobLd, line 101: can you rename `TryGetMemory(...)` into `TryGetBytesAsMemory(...)`? I think it makes the method function clearer

BobLd (changes requested):

BobLd:
@iamcarbon I've added more comments, thanks again for the hard work. Aside from that, all looks good to me

src/UglyToad.PdfPig/Content/BasePageFactory.cs
```diff
@@ -118,7 +118,7 @@ public abstract class BasePageFactory<TPage> : IPageFactory<TPage>
             }
             else if (DirectObjectFinder.TryGet<ArrayToken>(contents, PdfScanner, out var array))
             {
-                using var bytes = new ArrayPoolBufferWriter<byte>();
+                using var bytes = new ArrayPoolBufferWriter<byte>(1024 * 64);
```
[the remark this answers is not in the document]
BobLd: Update - the increase memory usage seems to not come from here, but my question stands. Thanks!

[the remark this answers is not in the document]
iamcarbon: I had set the initial buffer size in this case to cover the observed object sizes in the test suite. Most are in the 4KB range, but I saw some object sizes around 40KB. 

Otherwise, a 50KB segment has 8 additional copy operations, and array pool rentals, returns. 

e.g. 256 -> 512 > 1024 > 2048 -> 4096 -> 8192 -> 16384 -> 32768 -> 65536

We're likely to see a small regression on CPU time utilizing the pool, but much better steady state performance as we reduce the GC pressure.

iamcarbon:
> @iamcarbon I ran some benchmarks again following my previous comment. It seems the (alloc memory) performance decrease comes from your latest commit ([Remove unused StreamHelper methods](https://github.com/UglyToad/PdfPig/pull/815/commits/8314f411764d81ceb9f282214f63645b4b13b4c9))
> 
> memory2 (latest) image
> 
> memory2 (latest commit is `Eliminate allocation in BasePageFactory`) image

Will audit all our customer streams to see if there's a silent allocation somewhere. I changed our four WriteByte operations (which silently allocations on NET) to utilize a single Write operation, but I wouldn't expect that to allocate.

src/UglyToad.PdfPig/Util/PatternParser.cs
iamcarbon: I'll double check these are the terminals of the decode / filter pipe (in which case, this will be safe) and will update.

iamcarbon: good catch. this is safe! done.

BobLd:
This is the benchmark I ran

https://github.com/BobLd/Caly/blob/master/Caly.Benchmarks/OpenPdfBenchmark.cs

It does `GetImages()` which I think uses the 4 WriteBytes you replaced with a stackalloc

I agree with your remark though as I would not expect such a performance impact either...

iamcarbon:
I'll profile this, and continue to work on recreating / fixing this. This has my curiosity. 

Can you also confirm what framework you were testing against?

BobLd:
I was testing with dotnet8. I hope I didn't mess anything in the benchmark. Very curious too, let me know what you find

iamcarbon:
@BobLd 

Oddly, I'm seeing a consistent ~215MB allocated before and after the StreamHelper changes, from both the memory benchmark, and from a heap capture.

BEFORE StreamHelper changes
```
| Method | Mean     | Error   | StdDev   | Ratio | Gen0       | Gen1       | Gen2      | Allocated | Alloc Ratio |
|------- |---------:|--------:|---------:|------:|-----------:|-----------:|----------:|----------:|------------:|
| PdfPig | 506.0 ms | 9.93 ms | 11.44 ms |  1.00 | 28000.0000 | 14000.0000 | 7000.0000 | 215.79 MB |        1.00 |
```

AFTER StreamHelper changes
```
| Method | Mean     | Error    | StdDev   | Ratio | Gen0       | Gen1       | Gen2      | Allocated | Alloc Ratio |
|------- |---------:|---------:|---------:|------:|-----------:|-----------:|----------:|----------:|------------:|
| PdfPig | 528.0 ms | 10.27 ms | 12.99 ms |  1.00 | 35000.0000 | 20000.0000 | 8000.0000 | 215.78 MB |        1.00 |
```

When inspecting the heap, my guess is that the additional usage is from the ArrayPool rooting some new buffers. These would improve steady state performance at the cost of higher base memory usage. The ArrayPool would also get purged if there's memory pressure. 

Curious if we should try to optimize this, or work on reducing allocations elsewhere, and making further use of the pool to amortize the cost of keeping these new buffers around.

--- 

Some captures of the current heap usage:

Screenshot 2024-04-11 at 6 07 10 PM

Screenshot 2024-04-11 at 6 00 08 PM

--

Screenshot 2024-04-11 at 6 00 17 PM

BobLd:
@iamcarbon sounds good - let's leave that for another PR. 

Odd you don't see the same results though, I'll double check on my side

BobLd (approved):

PR #821, opened 2024-04-19
Reduce Allocations

iamcarbon:
This PR:

- Introduces a ValueStringBuilder
- Makes NumericTokenizer & PlainTextTokenizer thread-safe & reusable)
- Replaces the ListPool uses with ArrayPoolBufferWriter
- Seals various classes
- Eliminates various allocations
- Remove the unused (and allocating) FromOctalInt helper


iamcarbon:
@BobLd Ready for review.

src/UglyToad.PdfPig.Tokenization/Scanner/CoreTokenScanner.cs
```diff
@@ -13,11 +13,9 @@ public class CoreTokenScanner : ISeekableTokenScanner
         private static readonly CommentTokenizer CommentTokenizer = new CommentTokenizer();
         private static readonly HexTokenizer HexTokenizer = new HexTokenizer();
         private static readonly NameTokenizer NameTokenizer = new NameTokenizer();
+        private static readonly PlainTokenizer PlainTokenizer = new PlainTokenizer();
+        private static readonly NumericTokenizer NumericTokenizer = new NumericTokenizer();
```
iamcarbon, line 17: These are thread safe now and can be reused.

BobLd:
UglyToad.PdfPig.Fonts is missing the following:

```
<ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='net462' or '$(TargetFramework)'=='net471'">
	<PackageReference Include="System.Memory" Version="4.5.5" />
</ItemGroup>
```

BobLd:
Can you give more background about `ValueStringBuilder`? When you should and should not it be used over `StringBuilder`?

iamcarbon:
> Can you give more background about `ValueStringBuilder`? When you should and should not it be used over `StringBuilder`?

There's a few caveats to using ValueStringBuilder:

- it's a ref struct and needs to be disposed (by using "using", calling dispose, or calling ToString()). If used across functions, it needs to be passed by ref. If you miss the ref, it will create a copy of it's internal state, and can result in a double return.

- it can be slower for building large strings when a proper buffer size (hint) isn't provided. StringBuilder uses a linked list behind the scenes, while ValueStringBuilder double's its internal buffer and copies. 

- it doesn't support advanced StringBuilder uses like Replace or Insert.

With the caveats mentioned:

- it's allocation free, and allows you to use the backing buffer directly to avoid materializing a string.

- it can also be provided with stack memory for it's initial buffer, which avoids the overhead in using the array pool if the written bytes fit.

I'd make a case for using it nearly everywhere where the values are small or we are fairly confident in choosing the initial buffer size, and that doesn't make any use of advanced StringBuilder cases. It makes a nice dent in reducing allocations.

BobLd (approved):

PR #825, opened 2024-04-30
Improve code quality

iamcarbon:
- Simplifies the pragma statements
- Eliminates a few possible virtual calls
- Avoid encoding in a few more cases where we write known ASCII data

### DIFF
--- a/src/UglyToad.PdfPig/Graphics/Operations/OperationWriteHelper.cs
+++ b/src/UglyToad.PdfPig/Graphics/Operations/OperationWriteHelper.cs
@@ -9,7 +9,7 @@
 
     internal static class OperationWriteHelper
     {
-        private const byte Space = (byte)' ';
+        private const byte Whitespace = (byte)' ';
         private const byte NewLine = (byte)'\n';
 
         public static void WriteText(this Stream stream, string text, bool appendWhitespace = false)
@@ -36,7 +36,7 @@
 #endif
             if (appendWhitespace)
             {
-                stream.WriteByte(Space);
+                stream.WriteByte(Whitespace);
             }
         }
 
@@ -46,7 +46,7 @@
 
             if (appendWhitespace)
             {
-                stream.WriteByte(Space);
+                stream.WriteByte(Whitespace);
             }
         }
 
@@ -65,7 +65,7 @@
 
         public static void WriteWhiteSpace(this Stream stream)
         {
-            stream.WriteByte(Space);
+            stream.WriteByte(Whitespace);
         }
 
         public static void WriteNewLine(this Stream stream)
@@ -85,7 +85,7 @@
         public static void WriteNumberText(this Stream stream, int number, string text)
         {
             stream.WriteDouble(number);
-            stream.WriteByte(Space);
+            stream.WriteByte(Whitespace);
             stream.WriteText(text);
             stream.WriteNewLine();
         }
@@ -93,7 +93,7 @@
         public static void WriteNumberText(this Stream stream, int number, ReadOnlySpan<byte> asciiBytes)
         {
             stream.WriteDouble(number);
-            stream.WriteByte(Space);
+            stream.WriteByte(Whitespace);
             stream.WriteText(asciiBytes);
             stream.WriteNewLine();
         }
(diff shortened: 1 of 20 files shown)


iamcarbon:
@BobLd Ready for review. 

Best reviewed ingoring whitespace: https://github.com/UglyToad/PdfPig/pull/825/files?diff=unified&w=1

src/UglyToad.PdfPig/Graphics/Operations/OperationWriteHelper.cs
```diff
@@ -9,7 +9,7 @@
 
     internal static class OperationWriteHelper
     {
-        private const byte WhiteSpace = (byte)' ';
+        private const byte Space = (byte)' ';
```
BobLd, line 12: let's keep the name `WhiteSpace` for consistency

src/UglyToad.PdfPig/Writer/Fonts/ToUnicodeCMapBuilder.cs
```diff
@@ -1,117 +1,106 @@
 ﻿namespace UglyToad.PdfPig.Writer.Fonts
 {
+    using Graphics.Operations;
```
BobLd, line 3: system first

src/UglyToad.PdfPig/Writer/Fonts/ToUnicodeCMapBuilder.cs
```diff
@@ -1,117 +1,106 @@
...
     using Util;
 
     internal static class ToUnicodeCMapBuilder
     {
-        private const string BeginToken = "begin";
-        private const string BeginCMapToken = "begincmap";
-        private const string DefToken = "def";
-        private const string DictToken = "dict";
-        private const string FindResourceToken = "findresource";
+        private static ReadOnlySpan<byte> BeginToken => "begin"u8;
```
BobLd, line 12: `static readonly`

src/UglyToad.PdfPig/Writer/TokenWriter.cs
```diff
@@ -64,7 +64,7 @@ public class TokenWriter : ITokenWriter
 
         private static ReadOnlySpan<byte> TrueBytes => "true"u8;
 
-        private static readonly byte Whitespace = (byte)' ';
+        private const byte Space = (byte)' ';
```
BobLd, line 67: same here

BobLd (changes requested):

BobLd:
@iamcarbon can you fix the warning in the code too? (overrides in exceptions will need #if !NET I think)

src/UglyToad.PdfPig/Writer/Fonts/ToUnicodeCMapBuilder.cs
iamcarbon: This is the odd syntax they chose for statically initialized data, which is a method rather than a property (so we can't add readonly). It points to a fixed position in the assembly, which is readonly and immutable. It looks like it's allocating, but it's not.

https://vcsjones.dev/csharp-readonly-span-bytes-static/

NOTE, this only works for bytes on .NET versions before .NET 7.0.

iamcarbon:
@BobLd updated!

src/UglyToad.PdfPig/Writer/TokenWriter.cs
```diff
@@ -219,12 +219,11 @@ public void WriteToken(IToken token, Stream outputStream)
                      * n is a literal keyword identifying this as an in-use entry
                      * eol is a 2-character end-of-line sequence ('\r\n' or ' \n')
                      */
-                        var paddedOffset = OtherEncodings.StringAsLatin1Bytes(offset.Offset.ToString("D10", CultureInfo.InvariantCulture));
+                        var paddedOffset = Encoding.ASCII.GetBytes(offset.Offset.ToString("D10", CultureInfo.InvariantCulture));
```
BobLd, line 222: why changing that?

BobLd (changes requested):

src/UglyToad.PdfPig/Writer/TokenWriter.cs
iamcarbon: Is a bit faster to use the ASCII encoder on ASCII only digits (and we're guaranteed to only get back ASCII digits here). I had planned to optimize this on #NET later to encode directly to an ASCII byte[] buffer.

BobLd (approved):

PR #826, opened 2024-05-02
Add AndroidSystemFontLister

BobLd:


BobLd:
@EliotJones now good for review

src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs
```diff
@@ -89,6 +89,12 @@ static SystemFontFinder()
             {
                 lister = new LinuxSystemFontLister();
             }
+#if NET
```
EliotJones, line 92: presumably because it could be mono, or some other reason?

EliotJones (approved):
Thanks 🎉

src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs
BobLd: This was the less restrictive. The IsAndroid check does not exist for framework (obviously hehe)

PR #828, opened 2024-05-03
Cleanup using statements

iamcarbon:
This PR tidies up the using statements in the main library by:

- Importing System & System.Collections.Generic into the global scope
- Removing various unused statements
- Fixing misordered statements


iamcarbon:
@BobLd Ready for feedback. This should reduce the changes of statements being mis-ordered, and helps reduce code noise (particularly when working on a laptop).

BobLd:
@iamcarbon thx for that. I won't be merging this one, there are too many file changes.

What you did makes sense, but let's do these changes when the relevant file are modified

PR #829, opened 2024-05-05
Fix a compile error in the Examples project

Numpsy:
I tried building this whilst looking at something else, and it failed to compile with an error about ```TryGetBytes``` not existing.
Has it got out of step with the recent spanification changes?

### DIFF
--- a/examples/ExtractImages.cs
+++ b/examples/ExtractImages.cs
@@ -15,9 +15,9 @@
                 {
                     foreach (var image in page.GetImages())
                     {
-                        if (!image.TryGetBytes(out var b))
+                        if (!image.TryGetBytesAsMemory(out var b))
                         {
-                            b = image.RawBytes;
+                            b = image.RawMemory;
                         }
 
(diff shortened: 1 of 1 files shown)


Numpsy:
On a related note, the examples project is built as ```netcoreapp2.1``` - should it be updated to something newer to go with all the othre recent updates?

BobLd:
Thx a lot for that! Not sure how it didn't raise any issue when merging

> On a related note, the examples project is built as `netcoreapp2.1` - should it be updated to something newer to go with all the othre recent updates?

Not sure tbh, we could target a more recent framework indeed

BobLd (approved):

Numpsy:
> Not sure tbh, we could target a more recent framework indeed

I suppose that it does have the advantage of running against the netstandard2.0 build of the libraries, now that the unit tests are only targetting higher versions.

BobLd:
Yes exactly, let's leave it like that

Issue #790, opened 2024-03-09
ReflectionGraphicsStateOperationFactory and NativeAOT builds


Numpsy:
A follow up to #664 and a question / enquiry:

As mentioned in 664, ```ReflectionGraphicsStateOperationFactory``` is doing some things with reflection that can cause issues when using linker trimming or doing NativeAOT builds.

As I've been wanting to learn a bit about how SourceGenerators work, I wondered if if would be possible to generate the list of graphics operations with one, rather than having to enumerate all types in the library looking for classes that implement ```IGraphicsStateOperation``` at runtime.

So, I had a play and put together this: https://github.com/UglyToad/PdfPig/compare/master...Numpsy:PdfPig:source_gen

It's rather rough and glued together from generator samples, but the idea is that you can generate something like this at build time, and then just use the fixed collection of types with less runtime work:

![image](https://github.com/UglyToad/PdfPig/assets/1178570/6b5836b8-42e6-4087-a8ff-047773668677)

Does anyone have any thoughts on if the idea might be useful to take beyond a learning exercise?

Numpsy:
This has become a bit easier now, as it didn't work with .NET 4.5, but that's been removed by #795

jsiedentop:
I've tried using PdfPig with AOT and had trim warnings and an exception during runtime. Once I built my application with your fix, everything went fine. Is there a plan, to bring your code generation approach to main?

Numpsy:
It needs a bit more work to be production ready, but I can have another look if there's interest.

I've updated my branch with all the latest changes for now anyway.

jsiedentop:
I am definitely interested. Did a lot research with libraries in different languages and PdfPig would be my preferred choice. Unfortunately I am depending on AOT. I didn't realize code generation in C# was a thing now too, but it's nice that it's leading to more compatibility here. Currently I know little about c# code generation myself, so not sure, if I would be much help, but let me know if i can assist anywhere else.

Numpsy:
Now that everything has been updated to .NET 8 etc I'll have a go at doing a version using ```IIncrementalGenerator``` instead of ```ISourceGenerator``` as I gather that should be faster

BobLd:
@Numpsy I'm not very familiar with source generators but we could also take the approach to hard code stuf. This is the approach I took, see here https://github.com/BobLd/PdfPig/blob/caly-20240412/src/UglyToad.PdfPig/Graphics/ReflectionGraphicsStateOperationFactory.cs

[...]

PR #832, opened 2024-05-06
Use a fixed list of operations in ReflectionGraphicsStateOperationFac…

Numpsy:
…tory rather than searching via reflection.

The simple approach discussed in #790 , to just use a fix list of operations rather than doing anything clever.

It's a draft for comments, and because we might be able to add tests to ```GraphicsStateOperationTests``` to compare the fixed list with all the types it finds via ```GetOperationTypes()```

### DIFF
--- a/src/UglyToad.PdfPig/Graphics/ReflectionGraphicsStateOperationFactory.cs
+++ b/src/UglyToad.PdfPig/Graphics/ReflectionGraphicsStateOperationFactory.cs
@@ -1,5 +1,3 @@
-using UglyToad.PdfPig.Tokenization;
-
 namespace UglyToad.PdfPig.Graphics
 {
     using Operations;
@@ -9,6 +7,7 @@ namespace UglyToad.PdfPig.Graphics
     using Operations.InlineImages;
     using Operations.MarkedContent;
     using Operations.PathConstruction;
+    using Operations.PathPainting;
     using Operations.SpecialGraphicsState;
     using Operations.TextObjects;
     using Operations.TextPositioning;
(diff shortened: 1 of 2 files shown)


src/UglyToad.PdfPig/Graphics/ReflectionGraphicsStateOperationFactory.cs
```diff
@@ -20,37 +19,86 @@ namespace UglyToad.PdfPig.Graphics
...
-        private readonly IReadOnlyDictionary<string, Type> operations;
-
-        public ReflectionGraphicsStateOperationFactory()
-        {
-            var assemblyTypes = Assembly.GetAssembly(typeof(ReflectionGraphicsStateOperationFactory))!.GetTypes();
-
-            var result = new Dictionary<string, Type>();
-
-            foreach (var assemblyType in assemblyTypes)
+        private static readonly IReadOnlyDictionary<string, Type> operations =
```
Numpsy, line 25: As it stands, ```operations``` is instanced to the class, but as the contents is constant, is it ok for it be static instead?

Numpsy: Now that there is a .NET 8 build, it may or may not be useful to see if using a FrozenDictionary would have any benefits here? (just a thought, i don't have any real experience with using those yet)

BobLd: I think that'd be definitely a great use of frozen dictionary here indeed!

Numpsy: Updated to use FrozenDictionary in the .NET 8.0 build

Numpsy:
Updated with an attempt at a unit test for checking that the operations dictionary contains all the operations

BobLd (approved):
Thanks a lot for that @Numpsy, looks good to me with the tests!

PR #835, opened 2024-05-07
Update the dependencies in UglyToad.PdfPig.Package

Numpsy:
As it stands, the nightly build doesn't have any other deps than System.ValueTuple:

image

If I try to use that from a brand new / clean .NET Framework 4.7.1 project, I get this error:

image

So - I think the NuGet package should have a reference to System.Memory on TFMs other than .NET6/8 ?

Edit - Looks like it needs Microsoft.Bcl.HashCode as well

### DIFF
--- a/tools/UglyToad.PdfPig.Package/UglyToad.PdfPig.Package.csproj
+++ b/tools/UglyToad.PdfPig.Package/UglyToad.PdfPig.Package.csproj
@@ -28,8 +28,12 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


BobLd:
@Numpsy thx a lot for that, I'm really not sure how it was not caught by the unit tests...

BobLd (approved):

Numpsy:
> I'm really not sure how it was not caught by the unit tests...

I think it's because the unit tests are pulling in the library projects via ProjectReference so they always pull in transitive dependencies, but the nuget package is built with a different project that specifies dependencies explicitly?

BobLd:
mm yeah makes sense. I'll try to find a way to test the nuget packages directly

PR #843, opened 2024-06-02
Seal and update IFilters to return ReadOnlyMemory<byte>

BobLd:


BobLd:
@EliotJones now good to go

src/UglyToad.PdfPig.Tests/Filters/Ascii85FilterTests.cs
```diff
@@ -22,7 +22,7 @@ public void DecodesWikipediaExample()
 
             var result = filter.Decode(bytes, dictionary, 0);
 
-            var text = Encoding.ASCII.GetString(result);
+            var text = Encoding.ASCII.GetString(result.ToArray());
```
EliotJones, line 25: Presumably this is needed because this version https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=net-8.0#system-text-encoding-getstring(system-readonlyspan((system-byte))) isn't available in all the .NET versions we support?

EliotJones: Oh, span vs memory, all this stuff is new to me, any way to returns spans instead or they're only local to the method?

EliotJones (approved):

src/UglyToad.PdfPig.Tests/Filters/Ascii85FilterTests.cs
BobLd: We do have 2 extension methods for !NET version (in 2 different pdfpig) that are internal, but accessible from the test project. They create a conflict here as the test project cannot pick one of the 2. Hence the need to use non extensions methods

PR #852, opened 2024-06-23
Fix issues in Advanced Example in README.md

### LINKED ISSUE #851: Errors in examples on "readme.md" ?
Hello, 

I am discovering the PdfPig library, which seems very nice, but I found some confusing things in your examples, and especially the "Advanced Document Extraction" one of the main readme page:

- The `foreach (var word in page.GetWords())` seems totally useless (and monstruously increase execution time as it is around the whole treatment)
- The `pageBuilder.AddText(block.ReadingOrder.ToString(), 8, bbox, font);` does not work as is, since "bbox" parameter is apparently expected to be a point.

I could have done a PR to improve this but since I'm new with the library, I think it's better to explain it and let you fix it the best way.

Thank you

davebrokit:
Fixes #851

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,52 +84,52 @@ In this example a more advanced document extraction is performed. PdfDocumentBui
 
 
 ```cs
+//using UglyToad.PdfPig;
 //using UglyToad.PdfPig.DocumentLayoutAnalysis.PageSegmenter;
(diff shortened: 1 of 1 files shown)


AFract (approved):
Seems to fix at least problems raised in my issue #851

EliotJones (approved):
thanks!

Issue #855, opened 2024-06-25
Allow reading orders dectors to support any class that has a bounding box/PdfRectangle


davebrokit:
Currently the interface `IReadingOrderDetector` relies on `TextBlock` as a parameter. This limits it's use to the `TextBlock` class.

I propose adding an `IBoundingBox` interface

```cs
public interface IBoundingBox
{
    PdfRectangle BoundingBox { get; }
}
```

Then changing `IReadingOrderDector` interface and implementing classes to use `IBoundingBox` as it's parameter

Adding an overload that takes a `Func<T, PdfRectangle>` would allow the caller to specify any bounding box making the interface more useful.

Breaking changes: The `IReadingOrderDector` will instead return an `IReadOnlyList<T>` which will be the ordered results. This would mean `TextBlock.ReadingOrder` is not set which is a breaking change. But some code can be added that if type T is `TextBlock` then `ReadingOrder` is set

Happy to make the changes

BobLd:
@davebrokit I was thinking of doing similar, please go ahead and implement your idea. 

I did a similar interface for my project https://github.com/BobLd/Caly/blob/master/Caly.Pdf/Models/IPdfTextElement.cs feel free to reuse that or not.

I think the `Letter` class has a method instead of a property to get the bounding box. Might be a good opportunity to change that too (in my mind, the letters, text lines and text block should implement your interface, but please let me know what you think)

EliotJones:
Closing in line with #1095 will try and work out what the state of the PR is

PR #859, opened 2024-06-26
Allow reading orders dectors to support any class that has a bounding box/PdfRectangle

davebrokit:
- Introduce IBoundingBox interface

#855


src/UglyToad.PdfPig/Content/Letter.cs
```diff
@@ -1,171 +1,176 @@
-﻿namespace UglyToad.PdfPig.Content
-{
-    using Core;
-    using Graphics.Colors;
+﻿namespace UglyToad.PdfPig.Content
```
davebrokit, line 1: Not sure what happened here (Formating tabs/whitespace?). Only added interface and

```
        /// <summary>
        /// Gets the Bounding Box: The rectangle completely containing this object
        /// </summary>
        public PdfRectangle BoundingBox => GlyphRectangle;
```

src/UglyToad.PdfPig.DocumentLayoutAnalysis/ReadingOrderDetector/RenderingReadingOrderDetector.cs
```diff
@@ -17,14 +18,28 @@ public class RenderingReadingOrderDetector : IReadingOrderDetector
...
         /// Gets the blocks in reading order and sets the <see cref="TextBlock.ReadingOrder"/>.
         /// </summary>
-        /// <param name="textBlocks">The <see cref="TextBlock"/>s to order.</param>
-        public IEnumerable<TextBlock> Get(IReadOnlyList<TextBlock> textBlocks)
+        /// <param name="blocks">The <see cref="TextBlock"/>s to order.</param>
+        /// <returns>the orignal list if type is not <see cref="TextBlock"/></returns>
+        public IEnumerable<TBlock> Get<TBlock>(IEnumerable<TBlock> blocks)
+             where TBlock : IBoundingBox
+        {
+            if (typeof(TBlock) != typeof(TextBlock))
```
davebrokit, line 26: Should we throw here instead?

This reading order detector would only work with type TextBlock. Should we throw when the type is not that?

davebrokit:
Need to do some refactoring

Issue #844, opened 2024-06-07
XYLeaf.GetLines collect lines not robust enough


hjy1210:
I tryed to extract text from [history-2-3.pdf](https://hjy1210.github.io/pdf/history-2-3.pdf).

Part image of the history-2-3.pdf is
![history-2-3-item1](https://github.com/UglyToad/PdfPig/assets/7498709/1a3dd700-f910-4b81-a368-10e8bb213006)
It is a question item with item label "1." , item label followed by question body "新北市金瓜石的國際終戰和平紀念園區...".

But the extracted text are:
```
新北市金瓜石的國際終戰和平紀念園區，立有一面紀念牆，上面刻有第二次世界大戰時拘
1.
留在此的盟軍戰俘姓名與國籍。這些戰俘中，許多是紐西蘭及加拿大軍人。他們的國家並
未遭到日軍攻擊，卻仍參戰，其原因為何？
(A)受雇到海外作戰的傭兵 (B)由大英國協動員的軍隊
(C)反法西斯主義的志願軍 (D)聯合國維和部隊的成員
```
Note:  the item label `1.` is followed the first line `新北市金瓜石的國際終戰和平紀念園區，立有一面紀念牆，上面刻有第二次世界大戰時拘` of the question content. This phenomenon will cause trouble when searching text.

Following is my code:
```
            string sourcePdfPath = "history-2-3.pdf";
            string outputPdfPath = "history-2-3-pigboxed.pdf";
            string outputTextPath = "history-2-3-pigboxed.txt";

            using (var document = PdfDocument.Open(sourcePdfPath))
            {
                var builder = new PdfDocumentBuilder { };
                PdfDocumentBuilder.AddedFont font = builder.AddStandard14Font(Standard14Font.Helvetica);
                StringBuilder sb = new StringBuilder();
                for (var pageNumber = 1; pageNumber <= document.NumberOfPages; pageNumber++)
                {
                    var pageBuilder = builder.AddPage(document, pageNumber);
                    pageBuilder.SetStrokeColor(128, 128, 128);
                    var page = document.GetPage(pageNumber);

                    var letters = page.Letters; // no preprocessing

                    // 1. Extract words
                    var wordExtractor = NearestNeighbourWordExtractor.Instance;

                    var words = wordExtractor.GetWords(letters);

                    // 2. Segment page
                    var recursiveXYCut = new RecursiveXYCut(new RecursiveXYCut.RecursiveXYCutOptions()
                    {
[...]

PR #864, opened 2024-06-30
Introduce IBlock and ILettersBlock interfaces

davebrokit:
Part of #855

### DIFF
--- a/src/UglyToad.PdfPig.DocumentLayoutAnalysis/Export/PageXmlGeneralExporter.cs
+++ b/src/UglyToad.PdfPig.DocumentLayoutAnalysis/Export/PageXmlGeneralExporter.cs
@@ -1,0 +1,306 @@
+﻿namespace UglyToad.PdfPig.DocumentLayoutAnalysis.Export
+{
+    using Content;
+    using Core;
+    using DocumentLayoutAnalysis;
+    using Graphics.Colors;
+    using PAGE;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Xml;
+    using System.Xml.Serialization;
+
+    /// <summary>
+    /// PAGE-XML 2019-07-15 (XML) exporter for general case
+    /// This is a rewrite of <see cref="PageXmlTextExporter"/> to be simple and handle a general case of text, image
+    /// and custom implementer defined blocks
+    /// <para>See https://github.com/PRImA-Research-Lab/PAGE-XML </para>
+    /// </summary>
+    public class PageXmlGeneralExporter
+    {
+        private readonly double scale;
+        private string indentChar;
+        private int nextId;
+
+        /// <summary>
+        /// PAGE-XML 2019-07-15 (XML) exporter for general case
+        /// <para>See https://github.com/PRImA-Research-Lab/PAGE-XML </para>
+        /// </summary>
+        /// <param name="scale"></param>
+        /// <param name="indent"></param>
+        public PageXmlGeneralExporter(double scale = 1.0, string indent = "\t")
+        {
+            this.scale = scale;
+            indentChar = indent;
+        }
+
+        /// <summary>
+        /// Get the PAGE-XML (XML) string of the pages layout using the <see cref="IBlock"></see>'s as the page layout
+        /// </summary>
+        /// <param name="page">The Page</param>
+        /// <param name="blocks">Blocks to be exported</param>
+        /// <returns></returns>
+        public string Get(Page page, IEnumerable<IBlock> blocks)
+        {
+            PageXmlDocument pageXmlDocument = new PageXmlDocument()
+            {
+                Metadata = new PageXmlDocument.PageXmlMetadata()
+                {
+                    Created = DateTime.UtcNow,
+                    LastChange = DateTime.UtcNow,
+                    Creator = "PdfPig",
+                    Comments = "",
+                },
+                PcGtsId = "pc-" + page.GetHashCode()
+            };
+
+            var xmlPage = CreatePage(page.Height, page.Width, blocks);
+
+            pageXmlDocument.Page = xmlPage;
+
+            return Serialize(pageXmlDocument);
+        }
+
+        private PageXmlDocument.PageXmlPage CreatePage(double pageHeight, double pageWidth, IEnumerable<IBlock> blocks)
+        {
+            var pageXmlPage = new PageXmlDocument.PageXmlPage()
+            {
+                ImageFilename = "unknown",
+                ImageHeight = (int)Math.Round(pageHeight * scale),
+                ImageWidth = (int)Math.Round(pageWidth * scale),
+            };
+
+            var regions = blocks
+                .Select(b => ToRegion(b, pageWidth, pageHeight))
+                .Where(x => x != null).ToList();
+            pageXmlPage.Items = regions.ToArray();
+
+            var regionsOrder = regions.Select(x => x.Id);
+
+            var orderedRegions = GetOrderRegions(regionsOrder).ToArray();
+            pageXmlPage.ReadingOrder = new PageXmlDocument.PageXmlReadingOrder()
+            {
+                Item = new PageXmlDocument.PageXmlOrderedGroup()
+                {
+                    Items = orderedRegions,
+                    Id = "g" + NextId()
+                }
+            };
+
+            return pageXmlPage;
+        }
+
+        private IEnumerable<PageXmlDocument.PageXmlRegionRefIndexed> GetOrderRegions(IEnumerable<string> idOrder)
+        {
+            var index = 1;
+            foreach (var item in idOrder)
+            {
+                yield return new PageXmlDocument.PageXmlRegionRefIndexed()
+                {
+                    RegionRef = item,
+                    Index = index++
+                };
+            }
+        }
+
+        private PageXmlDocument.PageXmlRegion ToRegion(IBlock block, double pageWidth, double pageHeight)
+        {
+            if (block is TextBlock textblock)
+            {
+                return ToPageXmlTextRegion(textblock, pageWidth, pageHeight);
+            }
+
+            if (block is ILettersBlock blockOfLetters)
+            {
+                return ToPageXmlSimpleTextRegion(blockOfLetters.BoundingBox, blockOfLetters.Text, pageWidth, pageHeight);
+            }
+
+            if (block is IPdfImage imageBlock)
+            {
+                return ToImageRegion(imageBlock.BoundingBox, pageWidth, pageHeight);
+            }
+
+            // Default case
+            return ToPageXmlSimpleTextRegion(block.BoundingBox, block.ToString(), pageWidth, pageHeight);
+        }
+
+        private PageXmlDocument.PageXmlImageRegion ToImageRegion(PdfRectangle box, double pageWidth, double pageHeight)
+        {
+            return new PageXmlDocument.PageXmlImageRegion()
+            {
+                Coords = ToCoords(box, pageWidth, pageHeight),
+                Id = "r" + NextId(),
+            };
+        }
+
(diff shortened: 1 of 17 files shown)


src/UglyToad.PdfPig.DocumentLayoutAnalysis/Export/PageXmlGeneralExporter.cs
```diff
@@ -0,0 +1,306 @@
+﻿namespace UglyToad.PdfPig.DocumentLayoutAnalysis.Export
```
davebrokit, line 1: This is a copy of `PageXmlTextExporter` with changes to make it more general. With this you pass a list of IBlocks and it generates the PageXml for that.

src/UglyToad.PdfPig/Content/IBlock.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// The text of the block
+        /// </summary>
+        string Text { get; }
+
+        /// <summary>
+        /// Text orientation of the block.
+        /// </summary>
+        TextOrientation TextOrientation { get; }
+
+        /// <summary>
```
BobLd, line 31: What are you trying to achieve here by adding the letters collection? `Letters` collection should only be at `Word` level.

Having a collection of `Letters` at `Letter` level does not make sense and is counterintuitive... Similar can be said for `TextBlock` and `TextLine` classes.

src/UglyToad.PdfPig/Content/Letter.cs
```diff
@@ -7,12 +7,22 @@
...
     /// A glyph or combination of glyphs (characters) drawn by a PDF content stream.
     /// </summary>
-    public class Letter
-    {
+    public class Letter : ILettersBlock
+    {
+        /// <summary>
+        /// This letter as as List of Letters in order to implement ILettersBlock interface
+        /// </summary>
+        public IReadOnlyList<Letter> Letters => [this];
```
BobLd, line 15: see previous comment

BobLd (changes requested):

src/UglyToad.PdfPig/Content/IBlock.cs
davebrokit: Some of the reading orders use `TextSequence` to order words. This exposes the letters so we can get that.


"Having a collection of Letters at Letter level does not make sense and is counterintuitive... ". Goal there was to make Letters behave similar to other text blocks. But your comment makes sense so I'll change the `Letter` class

src/UglyToad.PdfPig.DocumentLayoutAnalysis/Export/PageXmlGeneralExporter.cs
davebrokit: Will Raise a separate PR and can discuss on that

src/UglyToad.PdfPig/Content/IBlock.cs
BobLd: If you want to be able to compute a `TextSequence` value at `TextBlock` and `TextLine`, exposing `IReadOnlyList<Letter>` is not the solution (as such, the `ILettersBlock` interface does not make sense to me). You need to find another way to compute the `TextSequence` value...

davebrokit: `ILettersBlock` interface then represents a collection of letters. Anyone can implement their own class that follows this contract as long as that class represents a collection of letters. E.g You can create a `Paragraph` class that implements this interface. It would make sense then to expose the `Letter`s that form the collection


Choices when implementing where:

1. Compute the TextSequence in the Reading Orders. This was currently done in multiple places. To reflect the IBoundingBox changes in the next PR this would mean introducing a generic function that checks type and if type is `TextBlock` `TextLine` or `Word` then return AverageTextSequence. This option was unnecessarily complex (generics can be painful) and limits the `RenderingReadingOrderDetector` to only work with the above classes. `UnsupervisedReadingOrderDetector` also has an option to use RendingOrder which would be limited
2. `ILettersBlock` with a field `AverageTextSequence`. This works but you force every implementer of `ILettersBlock` to implement this. Not necessarily useful for anything else but the ReadingOrderDetectors
3. `ILettersBlock` expose `IReadOnlyList<Letter>` This felt like the best solution as `RenderingReadingOrderDetector` and `UnsupervisedReadingOrderDetector` code with generics was simpler (if it has letters property then we know we can calculate the average text sequence)

`UnsupervisedReadingOrderDetector` looks like this with generics. `IBlock` to be 
```cs
public IEnumerable<TBlock> Get<TBlock>(IEnumerable<TBlock> blocks)
     where TBlock : IBoundingBox
{
    int readingOrder = 0;
    // Ordered by is a stable sort: if the keys of two elements are equal, the order of the elements is preserved 
    var ordered = blocks.OrderBy(b => GetAverageTextSequenceOrDefaultToZero(b));

    foreach (var block in textBlocks.OrderBy(b => AvgTextSequence(b)))
        if (typeof(TBlock) == typeof(TextBlock))
        {
            block.SetReadingOrder(readingOrder++);
            yield return block;
            return SetReadingOrder(ordered);
        }

    return ordered;
}

private double GetAverageTextSequenceOrDefaultToZero<TBlock>(TBlock block)
    where TBlock : IBoundingBox
{
    if (block is ILettersBlock textBlock)
    {
        return textBlock.Letters.Average(x => x.TextSequence);
    }

    return 0;
}
```

BobLd: - I think the approach you want to take is too generic. For example `UnsupervisedReadingOrderDetector` only makes sense at `TextBlock` level and should stay the same (see [paper](https://mgrani.github.io/media/papers/14kla.gra.jac.ea-Unsupervised-document-structure-analysis.pdf) is used at the time). This does not mean we could not have `ReadingOrderDetector` at other levels. I do agree with the need of a `IBoundingBox` interface though.

Nothing prevents you to do the following:

```csharp
public interface IReadingOrderDetector<T> where T: IBoundingBox
{
        IEnumerable<T> Get(IReadOnlyList<T> textBlocks);
}

[...]

public class RenderingReadingOrderDetector : IReadingOrderDetector<TextBlock> 
{
        [...]
}

[...]

public class UnsupervisedReadingOrderDetector : IReadingOrderDetector<TextBlock>
{
        [...]
}
```

One mistake (I believe) I did at the time was to set a reading order property at `TextBlock` level (`block.SetReadingOrder(readingOrder++);`). This property is in fact not needed and is redundant (`DefaultReadingOrderDetector` doesn't even sets it).

- Also, exposing these `Letter` collection everywhere means you are allocating _a lot_ of arrays, which is something we really want to avoid (see previous PRs about optimisation by @iamcarbon):
```csharp
words.SelectMany(w => w.Letters).ToList().AsReadOnly();
```

All that being said, I feel we moved too far away from the original problem raised in #844. In line with what was described in the issue, I too believe the issue is with 
> I guess this issue is due to: XYLeaf.GetLines use Words.Groupy(x => x.BoundingBox.Bottom) is not robust if words baselines/bottoms have some tiny differences.

I hope you won't take my feedback too harshely as you've already contributed greatly to the repo, but I'm not willing to merge the PR as is.

davebrokit: No issues with your feedback. It's a productive discussion as several ways to implement this feature. I had some of the questions you raised when implementing so happy to discuss them with someone rather than just me :)

The issue in question here is #855. #844 will be raised on a separate PR but depends on this for some of the refactoring I performed.

_Your comments:_
_Approach too generic_
a. `UnsupervisedReadingOrderDetector` -> 

- I added some tests at the `Word` level and it doesn't make sense.  But UnsupervisedReadingOrderDetector does make sense at the `TextLine` level, and would make sense if someone creates blocks that are a combination of `TextBlocks` and something else (Example is I've used it for my own custom class is a wrapper around a `TextBlock` or a `Table` from TabulaSharp which you wrote). So I would argue that it is generic enough to be used at any level apart from the word level (Any class that implements `IBoundingBox`)
- Secondly the interface should be generic enough that someone can contribute to the library their own reading detector orders for `Word`'s. For example a words in a line reading order detector. In the follow on PR I added something like that.


_Nothing prevents you to do the following:_
I tried that with the initial implementation. Following the argument in a) that is should be generic enough then we have to include the T parameter in the class definition otherwise we have multiple `UnsupervisedReadingOrderDetector` one for each type and if a user has their own custom type things get funky... . The problem I ran into here is that this breaks compatabilty with existing code using the library as you need add the T parameter when constructing:
```cs
public class RenderingReadingOrderDetector<T> : IReadingOrderDetector<T> 
{
        [...]
}

public class UnsupervisedReadingOrderDetector<T>: IReadingOrderDetector<T>
{
        [...]
}

var it = new RenderingReadingOrderDetector<TextBlock>()
```
Using a generic on the method keeps compatibilty with existing code.

c. SetReadingOrder -> I kept it in for compatibilty sake but happy to take it out.


 _Also, exposing these Letter collection everywhere means you are allocating a lot of arrays, which is something we really want to avoid (see previous PRs about optimisation by_
I'm happy to change it to a lazy property that computes when fetched. That would remove the performance issue

```
public IReadOnlyList<Letter> Letters => words.SelectMany(w => w.Letters).AsReadOnly();
```

davebrokit: @BobLd Just wanted to follow up on above

BobLd: @davebrokit I'm still here, I just don't have the time short term to focus on this area of PdfPig. I'll try to make time soon to have another look at how to address the points we discsussed above

EliotJones:
@BobLd are we waiting on any further changes for this PR?

BobLd:
@EliotJones I'd like to leave this PR open as the point addressed here are relevant. I won't have time short term to focus on this part of the library as i'd like to focus on bug fixes and implementing missing parts of the pdf specifications

davebrokit:
Hi @EliotJones @BobLd 

It'll be great to get some progress on this PR. I've responded to your comments @BobLd on the changes you requested:

Done:
- Split my big PR into 3 PRs
- change to `IBoundingBox`
- Exclude `PageXmlGeneralExporter` from PR

Todo (was waiting for response to my comment before making the change):
- Remove `IBoundingBox` interface from letter `Letter`
- Remove `SetReadingOrder`
- Change `Letters` property to lazy collection (depending on response to comment)

The contentious issue is should `Word`, `TextLine`, `TextBlock` implement the new `ILettersBlock` interface and the reading orders use the `IBoundingBox` in a future PR. I've made my arguments why I think they should and I'm waiting on a response. I can then make the necessary changes and hopefully close this PR after that.


It'll be great to get a response as this is functionality I'm using and the PR has been hanging around for 3 months 😞

I also have 2 follow on PRs that are dependent on this: One to address the line segementer issue raised in #844 and one to change the Reading order detectors that finish off this issue #855. The code has been written for them, but I haven't raised them yet as changes may need to be made following the outcome of this discussion.

EliotJones:
Thanks for your patience here @davebrokit. I'd suggest the following changes to move this forward (apologies if there is some point here already addressed in the discussion, I haven't read the full thing):

- Keep the `Bounds` property on the image classes so this isn't a major breaking API change, you can have both `BoundingBox` and `Bounds` where `BoundingBox` just returns `Bounds`, or is potentially even an explicit interface property, e.g. `IBoundBox.BoundingBox => Bounds`, or whatever the exact syntax is.
- I think I'd lean to not exposing the `Letters` collection for now, I'd rather that complexity and cost lives in the specific reading order detectors than every consumer has to pay the performance cost of evaluating these collections. I'm wondering if we keep the `ILettersBlock` interface but the reading order detectors have their own `wrapper` classes for related classes. However since `TextLine` and `TextBlock` live in document layout analysis namespace I can't remember what consumers cause these types to be constructed. But since it's a contentious change I'd rather we isolate it to the last-possible consumer, it's better to get this in and refine than have it stuck in this state, so this feels like a compromise no one is happy with, the best kind.

PR #877, opened 2024-08-08
Get correct text orientation when base line points are equal and fix #741

### LINKED ISSUE #741: Wrong Oritentation of letter
I have following Problem i have a 2d technical drawing where text is written in every direction.
The GetWords() Method with NearestNeighbourWordExtractor works fine for me except for this example.
grafik
In the Image you can see a part of the PDF.
Where light blue is the word box.
Dark blue the letter box and green the Location.
My Problem now is that the Letter has the TextOrientation Horizontal which leads to a wrongly drawn Text box for it and maybe that the 7 and the 9 cant find each other with nearest neighbour.
I have tried to create a pdf which has the same problems but i couldnt get it to work.
Because there is a nda i cant share the file, but maybe you could point me in the right direction to find the problem and maybe find a solution for it

BobLd:


src/UglyToad.PdfPig/Content/Letter.cs
```diff
@@ -160,6 +172,35 @@ private TextOrientation GetTextOrientation()
...
+
+                case 180:
+                case -180:
+                    return TextOrientation.Rotate180;
+
+                case 90:
+                    return TextOrientation.Rotate270;
+            }
+
+            throw new Exception($"Could not find TextOrientation for rotation '{rotation}'.");
```
EliotJones, line 201: Not sure I follow this change but is there a risk that e.g. 270 throws, should we return `Other` here instead of throwing? Or 270 is just handled by the 90/-90 cases? Should we check that the value is divisible by 90 cleanly?

EliotJones (approved):

src/UglyToad.PdfPig/Content/Letter.cs
BobLd: Hi @EliotJones, apologies I completely missed your comment here...

You are 100%, this needs to be handled differently. I will push a fix shorlty

PR #879, opened 2024-08-22
Handle null token in DirectObjectFinder, handle null state in SetNamedGraphicsState(), add and test and fix #874

### LINKED ISSUE #874: "Object reference not set to an instance of an object."
## STR

`PdfDocument.Open(pdfBytes)` with the some PDF file. As it contains sensitive data, i unfortunately cannot attach it here and I was unfortunately unable to create a minimal example, but some hints:
* it was a quite complex one with images and some header etc. (not much text though)
* contains some comments shown on hover (but not with the Acrobat icon it usually shows)
* it was last edited with Firefox and some notes and drawings were added
* it was last edited at 2023-08/2023-09

Much like this and I tried to reproduce it with this example, but **it does not work**:
grafik

Thus, i only attach this image, because with the PDF I've created it is _not reproducible_.

### What happens

```
System.NullReferenceException
  HResult=0x80004003
  Nachricht = Object reference not set to an instance of an object.
  Quelle = UglyToad.PdfPig
  Stapelüberwachung:
   bei UglyToad.PdfPig.PdfExtensions.TryGet[T](DictionaryToken dictionary, NameToken name, IPdfTokenScanner tokenScanner, T& token)
```

Apparently, this is the line of failure:
https://github.com/UglyToad/PdfPig/blob/a99c0d25bfe76e4e7a919a42c52c99022ac769d3/src/UglyToad.PdfPig/PdfExtensions.cs#L24

### What should happen

At least `PdfDocumentFormatException` if you consider the file invalid.

However, IMHO, the file is valid an can be opened with both Adobe Acrobat Reader and Firefox. Thus, actually parsing it would be good.

Also, when **opening it with Adobe Acrobat Reader and re-saving it**, it can be parsed!

## System

PDFPig 0.1.8
reproducible on Windows 10

Interne Referenz: 2118

BobLd:


BobLd:
@EliotJones tests now fixed - good to review

jmjohnson05:
We just waiting for approval on this PR?

BobLd:
Yes we are, hopefully this will be reviewed this weekend

EliotJones (approved):

PR #886, opened 2024-08-30
Support trailer info to be a dictionary instead of an indirect reference

### LINKED ISSUE #884: Failure to extract /Info data from the trailer when it is not an indirect reference
Hi,
I encountered a pdf with the following trailer:
```
trailer
<</Size 7/Info<</Producer(SumatraPDF 3.2)>>/Root 1 0 R>>
```
The code currently sticks to the norm stating that it must be an indirect reference.
| Key | Type | value |
| ----- | -- | - |
| Info | dictionary | (Optional; ▶️ shall be an indirect reference ◀️ ) The document’s information dictionary. As described in 14.3.3, "Document information dictionary", this method for specifying document metadata has been deprecated in PDF 2.0 and should therefore only be used to encode information that is stated as required elsewhere in this part of ISO 32000 |

However, (when parsed in Lenient mode ?) it could be great to handle the case where the data is an direct object.

Greybird:
This hould fix #884

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Integration/DocumentInformationTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Integration/DocumentInformationTests.cs
@@ -96,5 +96,23 @@
                 Assert.Equal("D:20190306232856Z00'00'", information.CreationDate);
             }
         }
+
+        [Fact]
+        public void CanReadDocumentInfromationDirectory()
+        {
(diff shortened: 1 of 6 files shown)


BobLd (approved):

BobLd:
@Greybird thanks a lot for the PR and adding the tests

PR #897, opened 2024-09-01
Support not finding the Pages dictionary in lenient mode

### LINKED ISSUE #896: Indirect reference to null for /Pages throws
It can happen that the /Pages token is set to an indirect reference that cannot be found in the document (either because it is missing, or that the parsing failed).
In lenient mode, this should not break the document opening.

Kids objects could also point to non Page object, and this should not break the document opening in lenient mode.

Greybird:
Should fix #896


BobLd:
@Greybird the tests are failing, can you have a look

BobLd (approved):

PR #907, opened 2024-09-20
Fix #878

### LINKED ISSUE #878: PdfDocumentBuilder creates broken annotation elements when copying pages from specific source PDFs
I have an application that uses PdfPig to merge multiple PDF files. I'm using `PdfDocumentBuilder.AddPage` for this.
The users of the application have reported a few cases where the created (merged) PDF has missing elements (mostly annotations).

Depending on the PDF viewer application the elements are either:
- Completely missing (e.g. in Acrobat Reader)
- Visible, but rotated (e.g. in Wondershare PDFelement, Google Chrome and Microsoft Edge)
- Correctly shown (e.g. in Mozilla Firefox)

Sample input file: [Input.pdf](https://github.com/user-attachments/files/16669919/Input.pdf)

Sample code:
```cs
string inputFile = @"C:\Data\Input.pdf";
string outputFile = @"C:\Data\Output.pdf";

using var targetStream = File.Open(outputFile, FileMode.Create, FileAccess.Write);
using var outputDocument = new PdfDocumentBuilder(targetStream);
using var inputDocument = PdfDocument.Open(inputFile);

for (int i = 1; i <= inputDocument.NumberOfPages; i++)
{
    outputDocument.AddPage(inputDocument, i);
}
```

I've tested multiple versions of PdfPig and here are the results:
- 0.1.8: Created PDF looks **correct**
- 0.1.9-alpha-20240219-c2536: Created PDF is broken
- 0.1.9-alpha-20240312-845e3: Created PDF is broken
- 0.1.9-alpha-20240721-a99c0 (current version): Created PDF is broken

Input:
grafik

Output when shown in Acrobat Reader:
grafik

Ouput when shown in Microsoft Edge:
grafik

BobLd:
This is in order to fix #878 

Unfortunately some tests are now failing because we are now to able to assess to proper precision to use. We either can fix the test to make it pass, or we need to review the logic

```
 UglyToad.PdfPig.Tests.Parser.PageContentParserTests.CorrectlyWritesOperations
   Source: PageContentParserTests.cs line 34

Expected: ···"0 0 0 rg BT /F0 21.33 Tf 1 0 0 -1 0 140 T"···
Actual:     ···"0 0 0 rg BT /F0 21.32999999999999829 Tf 1"···
```

Leaving it here as a draft for the moment. @cremor feel free to have a look

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Writer/OperationWriteHelperTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Writer/OperationWriteHelperTests.cs
@@ -1,0 +1,236 @@
+﻿namespace UglyToad.PdfPig.Tests.Writer
+{
+    using System.Globalization;
+    using UglyToad.PdfPig.Graphics.Operations;
+
+    public class OperationWriteHelperTests
+    {
+        [Fact]
+        public void WriteDouble0()
+        {
+            using (var memStream = new MemoryStream())
+            {
+                OperationWriteHelper.WriteDouble(memStream, 0);
+
+                // Read Test
+                memStream.Position = 0;
+                using (var streamReader = new StreamReader(memStream))
+                {
+                    var line = streamReader.ReadToEnd();
+                    Assert.Equal("0", line);
+                }
+            }
+        }
+
+        [Fact]
+        public void WriteDouble5()
+        {
+            using (var memStream = new MemoryStream())
+            {
+                OperationWriteHelper.WriteDouble(memStream, 5);
+
+                // Read Test
+                memStream.Position = 0;
+                using (var streamReader = new StreamReader(memStream))
+                {
+                    var line = streamReader.ReadToEnd();
+                    Assert.Equal("5", line);
+                }
+            }
+        }
+
+        [Fact]
+        public void WriteDoubleMinus5()
+        {
+            using (var memStream = new MemoryStream())
+            {
+                OperationWriteHelper.WriteDouble(memStream, -5);
+
+                // Read Test
+                memStream.Position = 0;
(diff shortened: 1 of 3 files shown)


cremor:
Do you know how the expected value of 21.33 was produced back then? Which data type and format specifier was used? What is the exact value that is formatted, and is is read somewhere or calculated?

If we would change the F17 format specifier to F14 then this specific case would again be formatted as 21.33. Maybe even less precision is enough? How much precision does the PDF standard even support?

BobLd:
Using F14 sounds good to me. For background:
> Note: To represent real numbers, Acrobat 6 uses IEEE single-precision floating-point numbers,

image

cremor:
If I read this correctly then even F5 would be enough.

BobLd:
Yes you are right, that's the theoretical precision. I'm going for 9 (Microsoft recommends F7 for single precision.

I still have an issue with the change as the behaviour is not the same between Framework (some tests are failing) and dotnet8 (all tests are good)

cremor:
Yeah, there seems to be a difference in number formatting between .NET Framework and .NET 8:
.NET Framework: https://dotnetfiddle.net/DyH0RY
.NET 8: https://dotnetfiddle.net/1YjwJ8

But this only affects numbers with more than 15 digits. Is this actually a problem? How big would a PDF page need to be so that this could be a real coordinate/size?

Btw, the `WriteDouble3` test fails (in both runtimes) if the current culture isn't using a dot as a decimal separator. You should add `CultureInfo.InvariantCulture` to that `double.Parse` call in the test.

EliotJones (approved):

PR #935, opened 2024-11-12
Optimize cross reference object offset validation by avoiding nested loop

madelson:
I've been using PdfPig for text extraction on various files.

Recently, I found 1 file which was taking ~23s to open. Profiling showed that the hotspot was:
```
Dictionary<IndirectReference, long>.TryInsert(...)
at CrossReferenceObjectOffsetValidator.ValidateCrossReferenceOffsets(...)
at CrossReferenceParser.Parse(...)
```

After the changes here, the time to open the document drops to just 3s!

### DIFF
--- a/src/UglyToad.PdfPig.Core/IndirectReference.cs
+++ b/src/UglyToad.PdfPig.Core/IndirectReference.cs
@@ -6,7 +6,7 @@
     /// <summary>
     /// Used to uniquely identify and refer to objects in the PDF file.
     /// </summary>
-    public readonly struct IndirectReference
+    public readonly struct IndirectReference : IEquatable<IndirectReference>
     {
         /// <summary>
         /// A positive integer object number.
@@ -31,15 +31,15 @@
         }
 
         /// <inheritdoc />
+        public bool Equals(IndirectReference other)
+        {
+            return other.ObjectNumber == ObjectNumber && other.Generation == Generation;
+        }
+
+        /// <inheritdoc />
         public override bool Equals(object obj)
         {
-            if (obj is IndirectReference reference)
-            {
-                return reference.ObjectNumber == ObjectNumber
-                       && reference.Generation == Generation;
-            }
-
-            return false;
+            return obj is IndirectReference other && Equals(other);
         }
 
         /// <inheritdoc />
--- a/src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
+++ b/src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
@@ -24,11 +24,11 @@
                 return true;
             }
 
-            var builderOffsets = new Dictionary<IndirectReference, long>();
-            
             var bruteForceOffsets = BruteForceSearcher.GetObjectLocations(bytes);
             if (bruteForceOffsets.Count > 0)
             {
+                var builderOffsets = new Dictionary<IndirectReference, long>();
+
                 // find all object streams
                 foreach (var entry in crossReferenceTable.ObjectOffsets)
                 {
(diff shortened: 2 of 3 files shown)


src/UglyToad.PdfPig.Core/IndirectReference.cs
```diff
@@ -6,7 +6,7 @@
     /// <summary>
     /// Used to uniquely identify and refer to objects in the PDF file.
     /// </summary>
-    public readonly struct IndirectReference
+    public readonly struct IndirectReference : IEquatable<IndirectReference>
```
madelson, line 9: Seeing that dictionary insertion was the bottleneck initially made me think that Equals boxing might be the issue so I added IEquatable. However, this didn't make a noticeable difference in my particular case. Given how frequently this type is used as a dictionary key I think it is worth keeping (I notice that other PdfPig types do have this).

src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
```diff
@@ -24,11 +24,11 @@ public static bool ValidateCrossReferenceOffsets(IInputBytes bytes, CrossReferen
                 return true;
             }
 
-            var builderOffsets = new Dictionary<IndirectReference, long>();
-            
             var bruteForceOffsets = BruteForceSearcher.GetObjectLocations(bytes);
             if (bruteForceOffsets.Count > 0)
             {
+                var builderOffsets = new Dictionary<IndirectReference, long>();
```
madelson, line 30: Moved this inside the if since it wasn't referenced outside

src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
```diff
@@ -39,11 +39,11 @@ public static bool ValidateCrossReferenceOffsets(IInputBytes bytes, CrossReferen
                         // TODO: more validation of streams.
                         builderOffsets[entry.Key] = entry.Value;
                     }
+                }
 
-                    foreach (var item in bruteForceOffsets)
```
madelson, line 43: Moving this loo outside the foreach loop that starts on 33 was the real win. As far as I can tell, the two loops are independent, and we're still maintaining the order such that bruteForceOffsets are added after object offsets and thus could override them in builderOffsets.

I don't see a reason why this needs to be an inner loop.

src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
BobLd: What about doing `var builderOffsets = new Dictionary<IndirectReference, long>(bruteForceOffsets.Count);` here too?

BobLd:
@madelson thanks a lot for the PR! I've added a comment, let me know what you think.

By any chance, is there a way for you to share the problematic pdf (not required, would just be nice for benchmarking)?

src/UglyToad.PdfPig/Parser/FileStructure/CrossReferenceObjectOffsetValidator.cs
madelson: Agreed & updated!

madelson:
Thanks @BobLd !

> @madelson thanks a lot for the PR! I've added a comment, let me know what you think.

Agreed & addressed!

> By any chance, is there a way for you to share the problematic pdf (not required, would just be nice for benchmarking)?

I looked into it and unfortunately it's not one I can share. However, any large PDF that hits this codepath probably reproduces the issue. Not sure if any of those are in the benchmark set already. This was over 300 pages of slides.

BobLd:
looks great! thanks a lot for the contribution @madelson

BobLd (approved):

BobLd:
@madelson your changes will be made available as a nuget package overnight (pre released version), so you'll have that by tomorrow morning. 

The official release is not planned yet, but should be made available before end of year

BobLd:
@madelson see https://www.nuget.org/packages/PdfPig/0.1.10-alpha-20241114-8ca53

Issue #951, opened 2024-12-08
IsProbablyWhitespace for negative gaps


Kizaemon:
Hi, my PDF file has tables with "merged" cells, so there are cases where the gap is negative, but should be parsed as a whitespace in the function GetText.

Would you please consider adding a parsing Option to recognize the negative gaps as white spaces?

`WhitespaceSizeStatistics.IsProbablyWhitespace(**abs(gap)**, previous)`

BobLd:
@Kizaemon do you mind providing a sample document containing the issue? Or screenshot to illustrate? Thanks!

Kizaemon:
Sorry, I already submitted a pull request, which solves this issue. A sample screenshot is included there.
Thank you!

Kizaemon:
@BobLd Please see [#952]

PR #952, opened 2024-12-08
New GetText() option: NegativeGapAsWhitespace

Kizaemon:
When parsing PDF files with tables containing multiple lines in a cell or "merged" cells, the separate words can appear out of horizontal order. This option can better predict the spaces between the words.

### DIFF
--- a/src/UglyToad.PdfPig.DocumentLayoutAnalysis/TextExtractor/ContentOrderTextExtractor.cs
+++ b/src/UglyToad.PdfPig.DocumentLayoutAnalysis/TextExtractor/ContentOrderTextExtractor.cs
@@ -110,6 +110,11 @@
                     {
                         var gap = letter.StartBaseLine.X - previous.EndBaseLine.X;
 
+                        if (options.NegativeGapAsWhitespace)
+                        {
+                            gap = Math.Abs(gap);
+                        }
+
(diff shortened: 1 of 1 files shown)


Kizaemon:
Solves [#951]

Kizaemon:
This is a sample screenshot from the PDF, where the next word starts to the left of the previous word, making the negative gaps.

image

BobLd:
@Kizaemon thanks for sharing the information here.

Table extraction is generally tricky to do in pdf documents.

Can I ask you to look into [tabula sharp](https://github.com/BobLd/tabula-sharp) which is a package based on PdfPig but specialised in table extraction?

Kizaemon:
@BobLd Thank you for recommending tabula sharp, I will surely take a look at it.

As for this pull request, I still feel it could be useful to have an option to detect the negative gaps as whitespaces.
The current implementation does not do anything with negative letter distances.

BobLd:
@Kizaemon I tend to agree here, reviewing your code I started to wonder if we should not simply always use the absolute value. 

I'll give it a proper look later today

Kizaemon:
There is a possibility that this change (unconditional absolute value) can break the existing parsing patterns for the current users.
Therefore I have suggested to add an option for backwards compatibility.

You will know better the typical usage patterns by the existing users and may just absolute in all cases.

BobLd (approved):

PR #1069, opened 2025-07-02
Fix bug in PngFromPdfImageFactory where softmask is wrongly referenced.

orrest:
Before the fix, the extracted png is:
before

After:
after

And the pdf page where the image in: 
[example.pdf](https://github.com/user-attachments/files/21016637/example.pdf)

And I think that's because the mask used the image's bytes, lead to the unexpected extraced image.


BobLd (approved):

BobLd:
@orrest thanks a lot for the contribution, greatly appreciated!

Issue #926, opened 2024-10-18
PdfDocumentFormatException: Could not find the object number 750 0 with type DictionaryToken instead, it was found with type ObjectToken.


Sayayaya:
PDFPig can't read the attached PDF Document.

I have tried in 0.1.8 and 0.1.9

```
UglyToad.PdfPig.Core.PdfDocumentFormatException: Could not find the object number 750 0 with type DictionaryToken instead, it was found with type ObjectToken.
   at UglyToad.PdfPig.Parser.Parts.DirectObjectFinder.Get[T](IndirectReference reference, IPdfTokenScanner scanner)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.ParseTrailer(CrossReferenceTable crossReferenceTable, Boolean isLenientParsing, IPdfTokenScanner pdfTokenScanner, EncryptionDictionary& encryptionDictionary)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.OpenDocument(IInputBytes inputBytes, ISeekableTokenScanner scanner, InternalParsingOptions parsingOptions)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(IInputBytes inputBytes, ParsingOptions options)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(Byte[] fileBytes, ParsingOptions options)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(String filename, ParsingOptions options)
   at UglyToad.PdfPig.PdfDocument.Open(String filePath, ParsingOptions options)
```

```
using (PdfDocument document = PdfDocument.Open(filePath))
```

Where filePath is the path to the attached PDF document

[2408.16296v1.pdf](https://github.com/user-attachments/files/17437730/2408.16296v1.pdf)

ummm288:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

BobLd:
Fixed in #1072

PR #1072, opened 2025-07-05
Fix issue 926

EliotJones:
the file provided in issue https://github.com/UglyToad/PdfPig/issues/926 contains the following syntax
in pdf object streams:

```
% 750 0 obj
<< >>
```

currently we read the comment token and skip the rest
however this producer is writing nonsense to the stream.
comment tokens are only valid outside streams in pdf files
so we align to the behavior of pdfbox here by skipping the
entire line containing a comment inside a stream which fixes
parsing this file.


BobLd (approved):

BobLd:
@EliotJones great to see you around, thanks a lot for the fix!

Issue #782, opened 2024-02-17
Cannot convert a block padded by 4 'u' characters. (Parameter 'index')


praaline:
When parsing a 5-page PDF file, I am getting the error "Cannot convert a block padded by 4 'u' characters. (Parameter 'index')".

I used PDFill free tools to split the PDF into 5 separate files, and then the parsing worked fine (same code).

I am creating a PDFDocument with PdfDocument.Open(pdfFileName); then getting each page in a loop with document.GetPage(i + 1); and then looping over each word with page.GetWords(); I am also using the boundingBox property of each word.

Could it be related to https://github.com/UglyToad/PdfPig/issues/548 ?

BobLd:
@praaline would you be able to share the document that's causing the issue?

praaline:
@EliotJones yes, could you please give me an email address? We also need to know who will have access to it (let's exchange this information by email, though). Thank you

RikanoAlienaris:
Same issue here.

![image](https://github.com/UglyToad/PdfPig/assets/52498937/814ddbb9-2773-497d-8f3f-6e5aec416e90)

The excepcion is produced when i call GetPages() in order to read the document. I can provide a PDF if it's needed.

Thank you.

EliotJones:
This should be addressed in the latest nightly build. If not feel free to open a new issue.

PR #1076, opened 2025-07-08
skip single letter final blocks

EliotJones:
thanks for the reviews @BobLd, I'm not sure how long I'm back for but have a little motivation again at the moment

this change attempts a fix for #782 but without reference documents I'm not certain this is the actual fix

this aligns our decoder with the behavior of pdfbox and c (https://github.com/dcurrie/ascii85/tree/master) implementations where single character final blocks are ignored rather than being written. also makes the error more informative in case it is ever encountered again.

also adds more test cases for ASCII85.

it is possible this is hiding the problem and will move the error elsewhere but this matches the implementation behavior of the 2 reference implementations.

one other potential source for the error is if pdf supports '<~' as a start of data marker which i can't find in the spec but wikipedia says might be possible? https://en.wikipedia.org/wiki/Ascii85#Adobe_version however the source for this is a single implementation in a language I don't understand.

without documents to trigger the error i think this is the best fix for now

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Filters/Ascii85FilterTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Filters/Ascii85FilterTests.cs
@@ -34,10 +34,39 @@ O<DJ+*.@<*K0@<6L(Df-\0Ec5e;DffZ(EZee.Bl.9pF""AGXBPCsi + DGm >@3BB / F * &OCAfu2 
                 text);
         }
 
-        [Fact]
-        public void ReplacesZWithEmptyBytes()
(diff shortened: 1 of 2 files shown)


BobLd:
@RikanoAlienaris the issue you are facing should be fixed, but not in 0.1.10 

Can you try with the latest pre-release version? `0.1.11-alpha-20250715-016b7` should work

RikanoAlienaris:
You're absolutely right, I didn't realize it wasn't a stable version yet. It does look like the issue is resolved now. Sorry about that.

Issue #936, opened 2024-11-12
PdfDocumentBuilder creates broken file when copying pages from specific source PDFs


cremor:
I have an application that uses PdfPig to merge multiple PDF files. I'm using `PdfDocumentBuilder.AddPage` for this.
The users of the application have reported a case where the created (merged) PDF is invalid and contains broken/garbled text. When I open the created PDF in Acrobat Reader I even get an error message saying that the PDF page contains errors.

Sample input files:
- [Input.pdf](https://github.com/user-attachments/files/17712348/Input.pdf)
  According to its metadata this file was created with OpenOffice.
- [Input with Comments.pdf](https://github.com/user-attachments/files/17712349/Input.with.Comments.pdf)
  This is based on the same file as the first, but was edited with Wondershare PDFelement to add some comments as annotations.

Sample code:
```cs
string inputFile = @"C:\Data\Input.pdf";
string outputFile = @"C:\Data\Output.pdf";

using var targetStream = File.Open(outputFile, FileMode.Create, FileAccess.Write);
using var outputDocument = new PdfDocumentBuilder(targetStream);
using var inputDocument = PdfDocument.Open(inputFile);

for (int i = 1; i <= inputDocument.NumberOfPages; i++)
{
    outputDocument.AddPage(inputDocument, i);
}
```

I've tested the following versions of PdfPig, all are affected:
- 0.1.8
- 0.1.9
- 0.1.10
- 0.1.11-alpha-20250324-0754e

Input:
![grafik](https://github.com/user-attachments/assets/655ec6d3-c652-42e6-bfd7-39fba560fe00)

Output when shown in Acrobat Reader:
![grafik](https://github.com/user-attachments/assets/ba10f221-0369-4e76-adff-a22c9c7ea8e3)

Ouput when shown in Microsoft Edge:
![grafik](https://github.com/user-attachments/assets/25a2c9fb-8486-4d7f-95e0-2d54c93f74ec)

cremor:
This might be related to an embedded font. Sometimes the end users get an error message like "The embedded font "EVPYXN+NotoSerifCJKjp-Regular-Identity-H" coud not be loaded..." from Acrobat Reader. But that error isn't shown every time and I haven't seen it myself yet.

BobLd:
@cremor thanks for the added context. I'll try to have a look soon but any help here would be really appreciated 😄

cremor:
There are additional details which point to a problem with fonts:
[...]

PR #1077, opened 2025-07-10
fix copying of sub-dictionary when keys collide

EliotJones:
when copying from a ancestor node of a page's resource dictionary we were incorrectly writing nested nodes of e.g. /fonts to the root of the target dictionary, here we write to the intended target node correctly. this is a fix for #936


BobLd (approved):

BobLd:
@EliotJones I will let you close https://github.com/UglyToad/PdfPig/issues/936 if the issue is fully fixed

Issue #1078, opened 2025-07-10
The dictionary did not contain a number with the key /Size


jchristn:
```
Unhandled exception. UglyToad.PdfPig.Core.PdfDocumentFormatException: The dictionary did not contain a number with the key /Size. Dictionary way: .
   at UglyToad.PdfPig.Util.DictionaryTokenExtensions.GetInt(DictionaryToken dictionaryToken, NameToken name)
   at UglyToad.PdfPig.CrossReference.TrailerDictionary..ctor(DictionaryToken dictionary, Boolean isLenientParsing)
   at UglyToad.PdfPig.CrossReference.CrossReferenceTableBuilder.Build(Int64 firstCrossReferenceOffset, Int64 offsetCorrection, Boolean isLenientParsing, ILog log)
   at UglyToad.PdfPig.Parser.FileStructure.CrossReferenceParser.Parse(IInputBytes bytes, Boolean isLenientParsing, Int64 crossReferenceLocation, Int64 offsetCorrection, IPdfTokenScanner pdfScanner, ISeekableTokenScanner tokenScanner)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.OpenDocument(IInputBytes inputBytes, ISeekableTokenScanner scanner, ParsingOptions parsingOptions)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(IInputBytes inputBytes, ParsingOptions options)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(Byte[] fileBytes, ParsingOptions options)
   at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(String filename, ParsingOptions options)
   at UglyToad.PdfPig.PdfDocument.Open(String filePath, ParsingOptions options)
```
[JoyfulMemo_small.pdf](https://github.com/user-attachments/files/21165901/JoyfulMemo_small.pdf)

I encounter the above exception when processing the attached file.  

I tried with and without explicit `ParsingOptions`.

```
ParsingOptions options = new ParsingOptions
{
  SkipMissingFonts = true,
  UseLenientParsing = true
};

using (PdfDocument document = PdfDocument.Open(filename, options))
{
  ...
}
```

BobLd:
@jchristn the issue should now be fixed in the latest pre-release version (0.1.11-alpha-20250715-016b7 should work). I'll close for now but feel free to reopen if need be

PR #1079, opened 2025-07-11
fix off-by-one and optimize brute force xref search #1078

EliotJones:
when performing a brute force xref search we were ending up off-by-one, update the search to use a ring buffer to reduce seeking and fix xref detection

I want to do some more testing around this since I don't understand how the existing code ever worked, but I'll let the CI tests run overnight. I'm going away for a week now, potentially more.


BobLd:
@EliotJones do you want me to wait for you before merging? Or are you okay merging as is?

EliotJones:
@BobLd I want to run integration tests first and then try and isolate some test cases because it appears every brute force search would have been broken and I find that hard to believe.

EliotJones:
@BobLd this has test coverage now and should be ready to merge, I'll re-run the integration tests to ensure I didn't break anything while adding tests.

PR #1084, opened 2025-07-16
allow reading to continue if encountering an invalid surrogate pair

EliotJones:
as discussed in #895 investigating the corpus at:
https://digitalcorpora.s3.amazonaws.com/s3_browser.html#corpora/files/CC-MAIN-2021-31-PDF-UNTRUNCATED/zipfiles/0000-0999/

the input file 0000000.pdf contained a utf-16 surrogate pair in an input defined as usc2. the approach of various parsers varies here, adobe acrobat seems to hard crash, pdf js returns the same text we now parse, chrome parses the intended text (2 invalid characters and "ib exam"). we don't care too much about matching chrome exactly so doing the same as firefox is fine here

### DIFF
--- a/src/UglyToad.PdfPig/PdfFonts/Cmap/CMap.cs
+++ b/src/UglyToad.PdfPig/PdfFonts/Cmap/CMap.cs
@@ -147,6 +147,8 @@
 
         public int ReadCode(IInputBytes bytes, bool useLenientParsing)
         {
+            var myPosition = bytes.CurrentOffset;
+
             if (hasEmptyCodespace)
             {
                 var data = new byte[minCodeLength];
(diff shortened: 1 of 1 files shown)


EliotJones:
Other parsers identify that this is a flipped input string (LE string in BE context). The input is `0020d83dde31004900420020004500580041004d`, but if you flip the byte order for input pairs to LE you get:

```
20 00 → U+0020
3D D8 → U+3DD8 ✅ = 㷘 or similar
31 DE → U+31DE ✅ = ㄞ or similar
49 00 → U+0049 = I
42 00 → U+0042 = B
20 00 → U+0020 = space
45 00 → U+0045 = E
58 00 → U+0058 = X
41 00 → U+0041 = A
4D 00 → U+004D = M
```

BobLd (approved):

Issue #1071, opened 2025-07-02
ArgumentException: Invalid number of inputs


NikiforovAll:
When I try to:

```csharp
var page = document.GetPage(1);
```

![Image](https://github.com/user-attachments/assets/cbc1893e-c253-4a62-a4b0-e693b8210518)

Unfortunately, I am unable to share the file with you.

EliotJones:
Without access to the document it's going to be near-impossible to resolve this from the error alone. We'd need at least the colorspace stream information being read. To find the colorspaces you can open the file in Notepad++ and search for the regex `/N[ /]`. This should result in a dictionary (denoted by `<< /Key /Value /Key2 /Value /N 3 >>`) as well as the stream below it, for example:

```
523 0 obj<</Length 2575/Filter/FlateDecode/N 3/Alternate/DeviceRGB>>stream
H‰œ–yTSw<<-- SNIP -->>íœî(î´ï@ïÌðXðåñrñÿòŒóó§ô4ôÂõPõÞömöû÷Šøø¨ù8ùÇúWúçûwüü˜ý)ýºþKþÜÿmÿÿ ÷„óû
endstream
endobj
```

If you can pull out the content of any objects, e.g. between the obj and endobj markers, like `523 0 obj -> endobj` in the example above, as Hex then maybe we can help.

EliotJones:
Actually I just noticed the comment on your other issue linking to this old issue looks related to this problem rather than that one https://github.com/UglyToad/PdfPig/issues/1061

Can you try with the latest nightly to confirm if that resolved the issue?

EliotJones:
This should be fixed in the latest nightly build.

PR #1088, opened 2025-07-17
fix colorspace error when form xobject contains a transparency group

EliotJones:
when a form xobject contains a reference to a group xobject this can only be used to change attributes of the transparency imaging model. the old code was setting the main colorspaces incorrectly causing errors when the transparency component had a different number of channels. this was causing #1071 in addition to the failure in file 0000355.pdf of the test corpus

[0000355.pdf](https://github.com/user-attachments/files/21306781/0000355.pdf)


BobLd:
Note for myself: investigate the CS issue based on the documents

PR #1089, opened 2025-07-18
support bfrange having incorrect length in a cmap

EliotJones:
the corpus file 0001413.pdf has an off-by-one error in its count for cmap bfranges. here we exit early if an unexpected endbfrange operator is encountered early. this matches the pdfbox behavior:

https://github.com/apache/pdfbox/blob/067d56e4dbf5817158d1e910b9602d031ab2153d/fontbox/src/main/java/org/apache/fontbox/cmap/CMapParser.java#L373


EliotJones:
With this change we support parsing all 1000 documents of corpus set 0001 except the following.

Corrupt:
- 0001589.pdf
- 0001957.pdf (can parse with SkipMissingFonts)

BobLd (approved):

Issue #614, opened 2023-04-21
CopyFrom and AddPage both cause AddText to be written upside down


brianliebe:
I am trying to load an existing PDF into my program and write simple text (just the page number) to each page. I have come across a bug where if I use `AddPage` or `CopyFrom` to copy a page from the existing PDF into a newly created PDF, the text gets created upside down. If I just create a new blank page and write the text to it, it works fine.

My guess is that the rotation is not preserved or something, because it also writes it to a different corner (top-left) than the non-copied page (bottom-left). The text is also much smaller if I do the copy from the existing document. I cannot find any way to get the text to be displayed the correct orientation after copying a page into the new PDF.

Is there a way I can write directly onto the existing PDF without having to create a new PdfDocumentBuilder?

Using CopyFrom or AddPage:
<img width="551" alt="image" src="https://user-images.githubusercontent.com/25470007/233708761-08b6ed03-295c-499b-a3f5-38009823ab9e.png">

Just writing text, no copy:
<img width="550" alt="image" src="https://user-images.githubusercontent.com/25470007/233709510-db54ce60-d896-4217-8afe-a76c0297bb07.png">

```csharp
using (var document = PdfDocument.Open(pdfPath))
{
    var b1 = new PdfDocumentBuilder();
    var f1 = b1.AddStandard14Font(Standard14Font.Courier);
    var p1 = b1.AddPage(document, 1);
    p1.AddText("Test1", 12, new PdfPoint(20, 20), f1);

    var b2 = new PdfDocumentBuilder();
    var f2 = b2.AddStandard14Font(Standard14Font.Courier);
    var p2 = b2.AddPage(PageSize.A3);
    p2.AddText("Test2", 12, new PdfPoint(20, 20), f2);

    byte[] data1 = b1.Build();
    byte[] data2 = b2.Build();
    File.WriteAllBytes(@"test1.pdf", data1);
    File.WriteAllBytes(@"test2.pdf", data2);
}
```

tsurrdurr:
@brianliebe Hello, I tried replicating the issue with the code you attached, I tried with different docs, and every time I can't replicate the bug. The text ends up where it is supposed to be every time - bottom left. This might be an issue specific to the format of the source document you're using.

jason-katz:
@tsurrdurr I had this same issue and it was definitely something specific to the PDF itself. Would it help if I emailed you the offending PDF?

EliotJones:
[...]

Issue #1011, opened 2025-03-12
Add functionality to calculate existing global transformation matrix or add content absolutely


svengeance:
Yo! Thanks for the work you've done here, and @BobLd for taking over to give Elliot a well-deserved break.

Given the requirement of adding a QR code to the bottom-right corner of the page, I've ended up in a bit of a goose chase.
Stamping code:

```cs
PdfPageBuilder.AddedImage? pdfImage = null;

foreach (var originalPage in pdfDocument.GetPages())
{
	var copiedPage = db.AddPage(pdfDocument, originalPage.Number);
	var pdfImageLocation = new PdfRectangle(
		new PdfPoint(originalPage.Width - image.Width, 0),
		new PdfPoint(originalPage.Width, image.Height)
	);

	if (pdfImage is null)
		pdfImage = copiedPage.AddJpeg(imageStream, pdfImageLocation);
	else
		copiedPage.AddImage(pdfImage, pdfImageLocation); // Reuses the same image for all pages
}

return db.Build();
```

This worked for several PDFs, but failed on Chromium-produced PDFs. After spending an incredibly loathsome amount of time learning about how PDFs work, I understand the issue to be that these Chromium PDFs begin their content stream with a strange transformation matrix that inverts the Y-axis and applies a 25% scaling

`0.23999999 0 0 -0.23999999 0 792 cm`

This caused my QR placement to be completely out of sync for these documents.

------------------

My solution was to wrap the page copying into a push/pop of the graphics state:

```cs
var newPage = db.AddPage(originalPage.Width, originalPage.Height);

newPage.CurrentStream.Operations.Add(Push.Value);
newPage.CopyFrom(originalPage);
newPage.CurrentStream.Operations.Add(Pop.Value);

// Add QR as expected
```

This eliminates the global matrix applied at the top of the chromium PDF. And, thank the Gods, it works.

--------------------

My suggestion for this library is to add some functionality to either (or both!):

1: Provide a method that will iterate over the content stream and calculate the "current" transformation matrix
2: Enhance image-writing methods to allow them to be absolutely positioned. This could be implemented by pushing, inverting the current transformation matrix, writing the image, then popping to restore.

If I've gone about this the hard way and there's an easier way to do what I've done, I'm also all ears.

For your convenience I've attached 3 PDFs - the chromium PDF with a naive stamp before the push/pop fix, the same PDF _with_ my push/pop fix, and a PDF without a global transformation matrix (produced by PdfPig).

[...]

PR #1094, opened 2025-07-19
when writing content to an existing page inverse any global transform #614

EliotJones:
#614 

when adding a page to a builder from an existing document using either addpage or copyfrom methods the added page's content stream can contain a global transform matrix change that will subsequently change all the locations of any modifications made by the user. here whenever using an existing stream we apply the inverse of any active transformation matrix

there could be a bug here where if you use 'copy from' with a global transform active, we then apply the inverse, and you use 'copy from' again to the same destination page our inverse transform is now active and could potentially affect the second stream, but I don't think it will

this code isn't very clean but I think it will be sufficient for now until we can have a bigger refactor of this whole area

### DIFF
--- a/src/UglyToad.PdfPig/Writer/PdfContentTransformationReader.cs
+++ b/src/UglyToad.PdfPig/Writer/PdfContentTransformationReader.cs
@@ -1,0 +1,36 @@
+﻿namespace UglyToad.PdfPig.Writer;
+
+using Core;
+using Graphics.Operations;
+using Graphics.Operations.SpecialGraphicsState;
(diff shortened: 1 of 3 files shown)


EliotJones:
This is also the issue reported in #1011 though in that case @svengeance reports wrapping the added content with a `q/Q` push/pop operator pair which is maybe a 'safer' fix since inversing the matrix will always face the risk of floating point errors in getting back to exact zero. Let's try this for now and change it if needed.

BobLd (approved):

Issue #1096, opened 2025-07-20
Detect cyclic references during IndirectReferenceToken resolution


jan-sutter:
Hi,

I have a PDF, which unfortunately, I cannot provide as an example.
However, I could isolate the issue to the following XObject reference.
I removed everything in the stream data.

```
1576 0 obj
<<
/Length 0
/Type /XObject
/Subtype /Image
/Filter /FlateDecode
/BitsPerComponent 8
/Width 318
/Height 162
/ColorSpace /DeviceRGB
/DecodeParms 2383 0 R
/SMask 2384 0 R
/isEipStamp true
/Metadata 1576 0 R
>>
stream
endstream
endobj
```

The issue with the PDF is the /Metadata reference onto itself.
This results in a StackOverflowException in ResolveInternal and combination with Resolve in PdfExtensions.cs in UtglyToad.PdfPig.
I could locally resolve this issue by simply storing a list of visited ids and bail out if an object has already been visited during a resolution step.
However, I'm not very familiar with the internal PDF structure, so I can't say if my solution will prevent certain PDFs from being read correctly.
I will provide my solution as a PR as well.

I hope my issue description is sufficient enough.
If you need more information or if you want me to adjust my description, please let me know.

BobLd:
To replicate the issue:
[issue_1096.pdf](https://github.com/user-attachments/files/21336139/issue_1096.pdf)

PR #1097, opened 2025-07-20
check for cycles during indirect reference resolution

jan-sutter:
This PR "fixes" #1096

### DIFF
--- a/src/UglyToad.PdfPig/PdfExtensions.cs
+++ b/src/UglyToad.PdfPig/PdfExtensions.cs
@@ -66,7 +66,7 @@
             {
                 var filter = filters[i];
                 totalMaxEstSize *= GetEstimatedSizeMultiplier(filter);
-                
+
                 transform = filter.Decode(transform, stream.StreamDictionary, filterProvider, i);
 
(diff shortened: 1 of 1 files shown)


EliotJones (approved):
Thanks for the contribution! This looks like a reasonable fix, the only concern I have is whether reference generation number is needed in the lookup set too, but I think that's such an edge case as to be virtually impossible. Let's merge it and see if the integration tests all pass. I'll close the issue you raised alongside this since this PR should address it.

EliotJones:
Sorry @BobLd I jumped the gun here, I'm not sure a better fix to be honest without much more rework. We're already quite heavy on object creation so I imagine the performance impact in non-negligible but also not much beyond baseline. The problem with these recursive functions that can loop is you always need some way to break out of an invalid file, the other alternative would be pass a counter integer and just throw if it exceeds some max-val iterations but I think this approach allows us to parse the file rather than throwing which is a nicer fix.

BobLd:
100% agreed, you can either create a PR or ill push the change shortly

theolivenbaum:
Done: https://github.com/UglyToad/PdfPig/pull/1112

PR #1102, opened 2025-07-20
move file parsing to single-pass static methods

EliotJones:
for the file 0002973.pdf in the test corpus we need to completely overhaul how initial xref parsing is done since we need to locate the xref stream by brute-force and this is currently broken. i wanted to take this opportunity to change the logic to be more imperative and less like the pdfbox methods with instance data and classes.

currently the logic is split between the xref offset validator and parser methods and we call the validator logic twice, followed by brute-force searching again in the actual parser. we're going to move to a single method that performs the following steps:

1. find the first (from the end) occurrence of "startxref" and pull out the location in bytes. this will also support "startref" since some files in the wild have that
2. go to that offset if found and parse the chain of tables or streams by /prev reference
3. if any element in step 2 fails then we perform a single brute-force over the entire file and like pdfbox treat later in file-length xrefs as the ultimate arbiter of the object positions. while we do this we potentially can capture the actual object offsets since the xref positions are probably incorrect too.

the aim with this is to avoid as much seeking and re-reading of bytes as possible. while this won't technically be single-pass it gets us much closer. it also removes the more strict logic requiring a "startxref" token to exist and be valid, since we can repair this by brute-force anyway.

we will surface as much information as possible from the static method so that we could in future support an object explorer ui for pdfs.

this will also be more resilient to invalid xref formats with e.g. comment tokens or missing newlines.


src/UglyToad.PdfPig/Parser/FileStructure/FileHeaderOffset.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace UglyToad.PdfPig.Parser.FileStructure;
+
+/// <summary>
+/// How many bytes precede the "%PDF-" version header in the file. In some files this 'junk' can
+/// offset all following offset bytes.
+/// </summary>
+internal readonly struct FileHeaderOffset(int value)
```
BobLd, line 7: I think this would be a good use for `readonly **record** struct`, especially since you override the `Equals()` method. In order to avoid boxing, implementing `IEquatable<FileHeaderOffset>` might be necessary

src/UglyToad.PdfPig/Parser/FileStructure/IXrefSection.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public XrefOffsetCorrection CorrectionType { get; }
+
+    /// <summary>
+    /// If we had to apply a correction to locate this xref this is how many bytes from the original location we had to move.
+    /// </summary>
+    public long OffsetCorrection { get; }
+}
+
+internal enum XrefOffsetCorrection
```
BobLd, line 37: `XrefOffsetCorrection : byte` ?

src/UglyToad.PdfPig/Parser/FileStructure/FileHeaderOffset.cs
BobLd: See https://nietras.com/2021/06/14/csharp-10-record-struct/ for me details and benchmark

BobLd (changes requested):

BobLd:
@EliotJones don't hesitate if you don't agree with my feedback, or if you need help

src/UglyToad.PdfPig/Parser/FileStructure/FileHeaderOffset.cs
EliotJones: I don't think we can use this yet due to backwards compatibility, I get an error about init only properties being unsupported on-build.

BobLd: @Numpsy thanks for that, I believe we alredy use these polyfill in another parts of the lib. I'll take care of that in another PR

Issue #1082, opened 2025-07-14
Reference issue: Preserve annotations when copying pages in PdfDocumentBuilder


cremor:
I'm using `PdfDocumentBuilder.AddPage` to merge multiple PDF files. When multiple source PDF are digitally signed and contain visible signature annotations then the created (merged) PDF only shows the signature annotations from the first file that is added to the target PDF.

But this problem only happens when the target PDF is opened with Acrobat Reader. Other PDF viewers show all annotations just fine.
So this might also be a bug in Acrobat Reader. But I don't know nearly enough about PDFs to decide if the bug is in PdfPig or Acrobat. But I suspect the bug in PdfPig because if I use https://www.ilovepdf.com/merge_pdf to merge the PDFs then the output PDF looks correct in Acrobat Reader.

Note: This bug is only about the visible signature annotations. I know that the actual digital signature isn't (and can't be) copied.

Sample input files:
[Input1.pdf](https://github.com/user-attachments/files/21215767/Input1.pdf)
[Input2.pdf](https://github.com/user-attachments/files/21215772/Input2.pdf)
(Note: I know that the digital signatures of those files are invalid. This is because they are only test signatures. The bug also happens with valid signatures.)

Sample code:
```cs
string inputFile1 = @"C:\Data\Input1.pdf";
string inputFile2 = @"C:\Data\Input2.pdf";
string outputFile = @"C:\Data\Output.pdf";

using var targetStream = File.Open(outputFile, FileMode.Create, FileAccess.Write);
using var outputDocument = new PdfDocumentBuilder(targetStream);
using var inputDocument1 = PdfDocument.Open(inputFile1);
using var inputDocument2 = PdfDocument.Open(inputFile2);

outputDocument.AddPage(inputDocument1, 1);
outputDocument.AddPage(inputDocument2, 1);
```

I've tested the following versions of PdfPig, all are affected:
- 0.1.8
- 0.1.9 
- 0.1.10
- 0.1.11
- 0.1.12 (Note: No annotation block is visible on either page with this version. This additional regression was fixed again, while the main bug remains.)
- 0.1.13
- 0.1.14

Expected output:
Both the first and second page should have a visible annotation block.

Actual output:
If opened in Acrobat Reader: Only the first page has a visible annotation block. The second page instead shows the text ">>Sonstige<<", which would normally be overlapped by the annotation block.
[...]

PR #1103, opened 2025-07-24
make link copying more tolerant when adding page

EliotJones:
in #1082 and other issues relating to annotations we're running into constraints of the current model of building a pdf document. currently we skip all link type annotations, i think we can support copying of links where the link destination is outside the current document. however the more i look at this code the more i think we need a radical redesign of how document building is done because it has been pushed far beyond its current capabilities, i'll detail my thinking in the related pr in more detail


EliotJones:
@BobLd interested on your thoughts on the proposed redesign please since you've had a couple of years dealing with these issues too. Any thoughts on things to look out for?

PR #1140, opened 2025-09-01
File buffering read stream investigation

### LINKED ISSUE #1124: Exception thrown: 'Ugly Toad.PdfPig.Core.PdfDocumentFormatException' in Ugly Toad. PdfPig.dll: 'Could not find the startxref' for PdfPig:0.1.11
Hi,

I've upgraded to `PdfPig:0.1.11` and now all tests fail with:

Image

NB, the underlying stream is:  `Microsoft.AspNetCore.WebUtilities.FileBufferingReadStream`.

```csharp
_bufferedStream = new FileBufferingReadStream(stream, MemoryThreshold);
var pdf = PdfDocument.Open(_bufferedStream, new ParsingOptions { SkipMissingFonts = true, UseLenientParsing = true });
```

---

See an example PDF:
[example.pdf](https://github.com/user-attachments/files/21592330/example.pdf)

EliotJones:
fix #1124 where using a sliding window treated the stream length as exact. certain stream types can return the wrong length if the data isn't yet buffered


cremor:
How can this PR fix anything if no code was changed (only a test was added)?
And I'm not even sure if the new test is run in actions because of the `#if NET9_0_OR_GREATER` since build_and_test.yml doesn't specify .NET 9.

PR #1145, opened 2025-09-07
Draft release on master build

EliotJones:
This PR changes the actions as follows:

- Remove the -alpha-001 postfix from all project version numbers
- Pushes to master will draft a new release in GitHub
- Publishing the release from GitHub will build, test and push the full version and then commit an increment for the version in the project files as well and assembly and file version


BobLd:
@EliotJones the draft release was properly created on merge to `master`.

We could try publishing the release and check all goes well publish the next version

Issue #564, opened 2023-03-14
PDFDocument.Open issue in request process in web environment


AhmetAtasagun:
Hi

When I try to open the pdf file that I captured with the IFormFile object as PDFDocument.Open(path), I get the following error;

![NofileExistError](https://user-images.githubusercontent.com/47632025/224936383-5d2a3ca8-6dcd-4ad1-a59b-d567200ae3f3.jpeg)

Differently, when I try to open the pdf file that I captured with the IFormFile object as PDFDocument.Open(stream), I get the following error;

![notfindversionheader1](https://user-images.githubusercontent.com/47632025/224936457-90f478c2-3b84-4fe3-ad44-4c560aa8721a.jpeg)
![notfindversionheader2](https://user-images.githubusercontent.com/47632025/224936485-80477b34-35d3-4b6e-9780-a47335f83145.jpeg)

I was able to solve it like this;
It worked when I saved the PDF file from a location and read it from the location. With another method, it also worked when I converted the stream to an array as seen in the picture below.

![solved](https://user-images.githubusercontent.com/47632025/224937559-9fd61690-adfc-4491-b274-c99e56cf8f7c.jpeg)

I think there is a problem with PDFDocument.Open(...) stream and file path overloads.
For your information...

mvantzet:
Hi Ahmet,
In case #1 you are trying to open the file from disk using the Filename property, but there is no actual file found on disk; which I would also expect until you explicitly save it from the request.
In case #2 you are turning the filename into a byte[], not the actual file contents, so it is trying to parse the filename as a PDF.
In case #3 you are correctly using the stream.

EliotJones:
Thanks for the answers here @mvantzet and @BobLd. I think I'll leave streams as-is for now. While we could internally transform them to `MemoryStream`s internally to make it more user friendly `Stream` is intended to be the more 'advanced' API for people to work with inputs that may not necessarily fit in memory, I don't particularly want us to have to do the work of correcting the API usage but probably could be improved with documentation.

PR #1158, opened 2025-09-14
Handle non seekable stream by copying it into a memory stream and fix #1146

### LINKED ISSUE #1146: UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(Stream stream, ParsingOptions options) fails on non seekable stream
Hey All,

I tried to read a PDF from a HTTP stream with chunked encoding. That failed inside the above mentioned Open method:

```
Specified method is not supported.    
         at System.Net.Http.HttpBaseStream.get_Position()
         at UglyToad.PdfPig.Parser.PdfDocumentFactory.Open(Stream stream, ParsingOptions options)
         at UglyToad.PdfPig.PdfDocument.Open(Stream stream, ParsingOptions options)`
```

Digging further into the error, the whole thing failed in the first line of the PdfDocumentFactory's  internal static PdfDocument Open(Stream stream, ParsingOptions? options) constructor, while accessing the streams Position property without previously checking the CanSeek property.

I was able to work around the problem by changing my call to the open method like this:

```
            var pdfStream = await httpClientFactory
                    .CreateClient()
                    .GetStreamAsync(
                        options.PdfSourceUri,
                        cancellationToken
                    );
            byte[] content;
            using (var ms = new MemoryStream())
            {
                pdfStream.CopyTo(ms);
                ms.Position = 0;
                content = ms.ToArray();
            }
            var doc = PdfDocument.Open(content);
```
I'd rather suggest to change that inside the constructor, though. Changing it like this:

```
        internal static PdfDocument Open(Stream stream, ParsingOptions? options)
        {
            byte[] streamInput;
            using (var memoryStream = new MemoryStream())
            {
                stream.CopyTo(memoryStream);
                memoryStream.Position = 0;
                streamInput = memoryStream.ToArray();
            }
            return Open(streamInput, options);
        }
```
would probably achieve better results than the current code.

Regards, Jens

BobLd:


BobLd:
@EliotJones if you have time to review to check if you agree with the approach. It was decided at the time to not handle making the stream seekable.

See https://github.com/UglyToad/PdfPig/issues/564

EliotJones (approved):
Still not a huge fan of the idea, Streams are a pain in the butt to work with and so if you have a stream I'd prefer to leave the responsibility to the caller to work with it correctly, but this change will at least ensure they fall into the pit of success.

PR #1167, opened 2025-09-20
Add GetDescent() and GetAscent() to IFont and loose bounding box to letter

### LINKED ISSUE #1165: Feature Request: Support Loose Bounding Box for Letters
Currently, we have a tight bounding box `GlyphRectangle` for letters, which works well. However, in some cases, we may want a ‘loose’ box, becuase letters like `.` have very small tight boxes. Please see the image below: red boxes represent tight bounding boxes, and blue boxes represent loose boxes.

Image

For reference, `pdfium` provides two APIs for the two types of bounding boxes, while `pdfminer` defaults to using the loose box.

A loose bbox could be calculated by using the font's ascent and descent. 

An initial impl here: 

https://github.com/lihuanglx/PdfPig/blob/5fd4b58a828a5982fb750beaf451d1f4f9b63bfd/src/UglyToad.PdfPig/Graphics/ContentStreamProcessor.cs#L109

Also, having the loose bboxes with same height might be helpful for layout analysis? I’m not entirely sure.

BobLd:
Fix #1165 

@lihuanglx can you run some tests on your end to check it fits the feature request?

The approach is slightly different than yours, but I believe it achieves the goal

### DIFF
--- a/src/UglyToad.PdfPig.Core/TransformationMatrix.cs
+++ b/src/UglyToad.PdfPig.Core/TransformationMatrix.cs
@@ -224,9 +224,18 @@
         [Pure]
         public double TransformX(double x)
         {
-            var xt = A * x + C * 0 + E;
+            return A * x + E; // + C * 0 
+        }
 
-            return xt;
+
+        /// <summary>
+        /// Transform an Y coordinate using this transformation matrix.
+        /// </summary>
+        /// <param name="y">The Y coordinate.</param>
+        /// <returns>The transformed Y coordinate.</returns>
+        public double TransformY(double y)
+        {
+            return D * y + F;
         }
 
         /// <summary>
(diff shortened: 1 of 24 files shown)


lihuanglx:
@BobLd Thanks! It seems that the boxes are taller than mine in this file: 
[ic2.pdf](https://github.com/user-attachments/files/22449695/ic2.pdf)

My results:
image

This PR:
image

lihuanglx:
Here’s another file for testing. In this file, both my code and the code from this PR produce taller boxes.

[test_c.pdf](https://github.com/user-attachments/files/22451046/test_c.pdf)

BobLd:
@lihuanglx cacn you have a second look now? I've update to logic.

As always with fonts, nothing's easy

PR #1175, opened 2025-09-30
Guard against circular references in XRef tables/streams

ricflams:
- Detect and prevent an xref table/stream at a certain offset from being read twice; malformed xref tables with circular references could otherwise cause the table-reading to loop forever.
- Another approach could be to prevent TryReadTableAtOffset from changing the bytes' CurrentOffset to the lastObjPosition in its attempt to read a table (eg restore CurrentOffset after the attempt to read a Table) so the outer bytes-loop could continue its search through the entire bytes unaffected.


BobLd:
@ricflams I'm merging as-is, but feel free to later create a PR if you want to implement the other approach (if there's reasons for that, i.e. performance maybe)

ricflams:
@BobLd thanks, and I think this is just fine and probably the easiest approach to reason about.

PR #1186, opened 2025-10-14
Use zlib decode to properly use window size and checksum in flate filter

rhuijben:
Use the zlib data to verify that the compressed data is not corrupted. This avoids quite a few cases of processing known corrupted data.

For .Net 6+ we can use the ZlibStream to do this in a single step, but we need a bit more work to be Framework compatible. 
As we can't use the zlib framing now, we need to calculate the length of the uncompressed data ourselves. Fix a few related bugs to make this possible.


src/UglyToad.PdfPig.Tests/Parser/FileStructure/FirstPassParserTests.cs
```diff
@@ -109,10 +109,8 @@ 0000004385 00000 n
             %%EOF
             """;
 
-        if (Environment.NewLine == "\n")
```
rhuijben: This is a runtime check. Not what is in the source file.

rhuijben:
@BobLd I fixed the issues at the flatefilter level, but I think there is an issue in where the streams are detected. I think we should drop the final newline ("\r\n", "\r" or "\n") from the Memory object we hand down. 
Otherwise every filter type has to handle these exceptions.

BobLd:
@rhuijben thanks for the details. I was looking for documents (not necessarily in the current tests), that used to fail to open before adding `input.Slice(0, length);` and now succeed.

But anyway, let me know when good with the changes (no rush), I'll review them

src/UglyToad.PdfPig/Filters/FlateFilter.cs
```diff
@@ -55,29 +65,84 @@ public Memory<byte> Decode(Memory<byte> input, DictionaryToken streamDictionary,
...
+            {
+                using (var deflate = new DeflateStream(memoryStream, CompressionMode.Decompress))
+                using (var adlerStream = new Adler32ChecksumStream(deflate))
+                using (var output = new MemoryStream((int)(input.Length * 1.5)))
+                using (var f = PngPredictor.WrapPredictor(output, predictor, colors, bitsPerComponent, columns))
+                {
+                    adlerStream.CopyTo(f);
+                    f.Flush();
+
+                    var result = output.AsMemory();
```
BobLd: not in use

EliotJones:
So I'm just looking into this because it looks like it causes page content to be skipped here https://github.com/UglyToad/PdfPig/issues/1235

In PDFBox the default is to use a more tolerant parser for these corrupted streams: https://github.com/apache/pdfbox/blob/trunk/pdfbox/src/main/java/org/apache/pdfbox/filter/FlateFilterDecoderStream.java#L47 which references this issue https://issues.apache.org/jira/browse/PDFBOX-1232.

I've run the first 2500 files of the test corpus but didn't detect any additional files affected by this. If we revert to DeflateStream what would the consequences be? Perhaps we can use checksum validation for lenient mode off but keep the support for corrupted streams where lenient parsing is on (default)?

BobLd:
@EliotJones I'd say the priority is to be able to extract text. I'd need to have a look, but I'm not against reverting on my end if it breaks stuff...

EDIT: Would need to double check but might be related to https://github.com/UglyToad/PdfPig/issues/1243

PR #1192, opened 2025-10-23
Fix regression introduced in 3592fc8 where slicing the stream to the length breaks decoding

BobLd:
Unfortunately the regression was not caught with the tests in the project.

I caught that when running tests in the https://github.com/BobLd/PdfPig.Rendering.Skia project.

3 tests are now failing over there:

- UglyToad.PdfPig.Rendering.Skia.Tests.TestRendering.PdfPigSkiaTest(expectedImage: "AcroFormsBasicFields_1.png", pdfFile: "AcroFormsBasicFields.pdf", pageNumber: 1, scale: 2) Failed
- UglyToad.PdfPig.Rendering.Skia.Tests.TestRendering.PdfPigSkiaTest(expectedImage: "FontMatrix-concat_1.png", pdfFile: "FontMatrix-concat.pdf", pageNumber: 1, scale: 2) Failed
- UglyToad.PdfPig.Rendering.Skia.Tests.TestRendering.PdfPigSkiaTest(expectedImage: "caly-issues-58-2_1.png", pdfFile: "caly-issues-58-2.pdf", pageNumber: 1, scale: 2) Failed

image

I'm lacking time to add the tests to this projects, but wanted to at least fix that for now.

cc @rhuijben

### DIFF
--- a/src/UglyToad.PdfPig/PdfExtensions.cs
+++ b/src/UglyToad.PdfPig/PdfExtensions.cs
@@ -64,17 +64,6 @@
 
             var transform = stream.Data;
 
-            var length = stream.StreamDictionary.GetIntOrDefault(NameToken.Length, -1);
-
-            // If a length is available and it's smaller than the actual data length, use that. This trims whitespace (e.g. newlines) that might have been introduced during transport.
-            // And with that it handles some issues before individual filters have to deal with it.
-            //
-            // Do this before the first filter (to handle cases like multiple filters, etc).
-            if (length > 0 && length < transform.Length)
-            {
-                transform = transform.Slice(0, length);
-            }
-
             for (var i = 0; i < filters.Count; i++)
             {
                 var filter = filters[i];
@@ -102,17 +91,6 @@
             double totalMaxEstSize = stream.Data.Length * 100;
 
             var transform = stream.Data;
-
-            var length = stream.StreamDictionary.GetIntOrDefault(NameToken.Length, -1);
-
-            // If a length is available and it's smaller than the actual data length, use that. This trims whitespace (e.g. newlines) that might have been introduced during transport.
-            // And with that it handles some issues before individual filters have to deal with it.
-            //
-            // Do this before the first filter (to handle cases like multiple filters, etc).
-            if (length > 0 && length < transform.Length)
-            {
-                transform = transform.Slice(0, length);
-            }
 
             for (var i = 0; i < filters.Count; i++)
             {


EliotJones:
Thanks, I don't have time to review much anymore but I think it makes sense not to trust the declared `/Length` in most situations, people do things like putting the number of filters or length of the un-compressed data or some random number in there. It can improve the initial stream parse if it happens to be the right length but unfortunately you usually need to manually read the stream in every case.

rhuijben:
Do you need to drop the length check on both levels?

BobLd:
@rhuijben that makes sense yes, ill push a pr for tgat

BobLd:
@rhuijben my bad - I misunderstood your earlier comment.

> What I asked was if you see the problems at both levels?

I only saw issue at top level (in `PdfExtensions.Decode(...)`), but not at `FlateFilter` level. That being said, and in line with Eliot's message, I think we should not trust `/Length`, mainly for the reasons laid out by Eliot. Another reason I remove the trimming in the FlateFilter is that it seams to not be the correct place to trim - see below.

> You removed the length trimming in two places. And with that last commit in three.

Yes, I believe this is safer for now, in line with the above.

> I'm trying to find where it is safe to do the trimming. Just assuming that whitespace is part of the stream is not always the right thing to do... but just trimming isn't either, or the tests wouldn't have failed.

Sadly, I don't think it's going to be easy. Going back to the Pdf 2.0 specifications:
> 7.3.8.2 Stream extent
> Every stream dictionary shall have a Length entry that indicates how many bytes of the PDF file are used for the stream’ s data. (If the stream has a filter, Length shall be the number of bytes of _encoded_ data.)

From this, my understanding is that the check should be be done in the `PdfExtensions.Decode(...)` methods, before any decoding. Doing the check in the `FlateFilter` might happen after another decoding. I removed this trimming in this PR, as it was creating an issue. 

One possible approach would be to try decoding the stream first without any trimming. If this decoding fails, we trim the stream and try decoding again... Another approach could be to check for whitespaces. More than open to discuss, and my understanding of the specifications might be wrong.

As a side note, this is a prime example that pdf readers do not fully enforce the specs and are more lenient. This is also why PdfPig has a lenient mode.

Below are the documents that were failing before this PR:
[AcroFormsBasicFields.pdf](https://github.com/user-attachments/files/23243829/AcroFormsBasicFields.pdf)
[FontMatrix-concat.pdf](https://github.com/user-attachments/files/23243831/FontMatrix-concat.pdf)
[caly-issues-58-2.pdf](https://github.com/user-attachments/files/23243833/caly-issues-58-2.pdf)

PR #1203, opened 2025-11-07
Bugs/revert e11dc6b

BobLd:
@rhuijben I'm very sorry to have to revert the commit you made, given the work done here.

There is a regression with the code change and I was not able to find the exact root cause.

The issue was originally found here: https://github.com/BobLd/PdfPig.Rendering.Skia/pull/51

Some FlatDecode images are not decoded properly anymore. I have added a test that fails with the commit being reverted. I hope this will help in the debugging


rhuijben:
Busy with a few other projects but will get back to this. WIP is in my fork repository if anybody wants to take a look earlier.

Nice work in that Skia repository. It does exactly what we need at $dayjob, but currently to heavy to integrate it in our in house project flow.

Issue #704, opened 2023-09-25
Reference issue: Add hyperlinks in PdfDocumentBuilder


Zafoue:
I need to add Hyperlinks in a pdf  does pdfPig support it ?

ClashOfStyles:
I have the same question.

EliotJones:
No, PdfPig does not currently support adding Hyperlinks or any other Annotation types.

BobLd:
fix by #1232

PR #1232, opened 2025-12-25
Add Links to Pdf Generation

ochsnerd:
Allows adding links while generating a Pdf.

The `PdfDocumentBuilder` already supports writing Link-Annotation, but the `PdfPageBuilder` did not yet have the appropriate Methods.

I created `LinkAnnotation` to make the specification of a link more ergonomic. The `LinkAnnotation` does not inherit from `Annotation` since supporting the whole breadth of Pdf-Annotations while writing the Pdf was too much work. Additionally, the design of the `Annotation`-class is more suited to parsing and does not work well when writing a Pdf.


ochsnerd:
This PR would solve #704 and its duplicates. Is there something I can do help it move along?

PR #1246, opened 2026-02-05
Make extended graphics states stacked too

PsykerUdot:
Paragraph 8.10.2 of PDF Spec, Table 95 states that Form XObject own resources _"shall not be promoted to the outer content stream's resource dictionary"_.

This fix makes Extended Graphics State stacked, allowing some PDF documents to be rendered more correctly.

This fix doesn't touch any other aspects regarding to the aforementioned PDF spec item.

Required for https://github.com/BobLd/PdfPig.Rendering.Skia/pull/64


BobLd:
@PsykerUdot thanks a lot for the PR! Would have document you could share that are better rendered with you fix? If you can, add tests here. I might add tests in the skia rendering project too

Issue #1235, opened 2026-01-10
Text not found in watermarked pdf


kobruleht:
Trying to exttract text from attached pdf using using:

```
        using PdfDocument document = PdfDocument.Open(stream,SkiaRenderingParsingOptions.Instance);
        document.AddSkiaPageFactory();
   foreach(Page page in document.GetPages())
     Console.WriteLine( page.GetWords().Count() );  // outputs 0
```

Returns no words.

Using  UglyToad.PdfPig.Rendering.Skia

`            using MemoryStream ms = document.GetPageAsPng(pnr,5,100);
`
returns empty image.

Text can copied when PDF is opened in Chrome browser.

[iizieileamidagi ARVE_2745540212.pdf](https://github.com/user-attachments/files/24546850/iizieileamidagi.ARVE_2745540212.pdf)

PR #1254, opened 2026-02-15
revert flate decode handling to more lenient processing

EliotJones:
As discussed here https://github.com/UglyToad/PdfPig/pull/1186#issuecomment-3904911743

the change to use zlib/a adler checksum verification flow meant that invalid flate streams would not be decoded correctly. this caused issues for files that included invalid/missing checksums. this reverts the processing to the old approach for files like #1235

I'd like to check with @rhuijben before merging though to see if I'm missing the main reasons for the initial change, but in the meantime we'll run the tests on this PR to check what it would break.

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Parser/PageContentParserTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Parser/PageContentParserTests.cs
@@ -216,6 +216,68 @@ l";
             Assert.NotEmpty(result);
         }
 
+        [Fact]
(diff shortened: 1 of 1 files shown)


BobLd:
@EliotJones thanks a lot for that! Let's merge as it is and for performance impact, `[UglyToad.PdfPig.Benchmarks](https://github.com/UglyToad/PdfPig/tree/master/tools/UglyToad.PdfPig.Benchmarks)` could be used

PR #1260, opened 2026-02-22
Fix Benchmarks solution and add BruteForceBenchmarks

BobLd:
@EliotJones  No major performance downgrade in term of allocation, with a performance boost in term of speed. Is this what you were expecting?

I selected documents where `BruteForceSearcher.GetObjectLocations()` is used. We might want to benchmark with other documents too.

image

where `Latest` is `0.1.14-alpha-20260219-adb57` and `Local` is 9c0d6893e02edd39e855a03ab34297261af94740


EliotJones:
I'm not sure what could cause a speed boost to be honest but thanks for checking 🙌

Issue #1171, opened 2025-09-26
change request: give me an option to not load image contents into memory when calling PdfDocument.GetPages


HertBp:
First off, thank you for all the hard work.
Then, the problem I am encountering is the following:
we use azure batch with cheap nodes to process pdf's in bulk, these cheap nodes have low RAM.  Usually this works fine. However some publishers insist on cramming their pdf's with high resolution images. I am not interested in the contents of these images, only their size and location. But it looks like this contents is loaded into RAM anyway when I call PdfDocument.GetPages. This not only takes a lot of CPU, but also causes SystemOutOMemory exceptions which means we cannot parse the Pdf (and just wasted a bunch of processing time).
The main problem is publishers putting these stupidly detailed pictures in the pdf, but we do not have any power over them to stop them doing this so if I could tell PdfPig not to load the pictures into memory, just their PdfRectangle it would save a huge amount of money.

edit: if this already exists, please point me to it, I could not find it

Numpsy:
> edit: if this already exists, please point me to it, I could not find it

There's some discussion about having flags to disable some of the processing over in https://github.com/UglyToad/PdfPig/issues/980#issuecomment-3092634566

PR #1280, opened 2026-04-15
Add EagerlyLoadImageBytes parsing option

EdwardBlair:
Adds a ParsingOptions.EagerlyLoadImageBytes flag (defaults to true so nothing changes for existing users). When set to false, image byte data is stripped during page processing but all the metadata like width, height, bounding box, color space etc is still available. This should help with the OOM issues in #1171 where large images blow up memory on low-RAM machines, it does not prevent the image bytes from being read into memory but does flag them for GC.

Also makes XObjectImage.RawMemory lazy so even with the default settings, enumerating images for metadata doesn't force the bytes to be copied around.

This is a stop gap in lieu of a larger change to lazy the stream.

Line endings were mixed so I've normalised them. I will do a separate PR to normalise them across the entire repo as required. Git should handle this transparently.

> [!WARNING]
> Please view this with ?w=1 otherwise you will get spammed with line ending changes
> https://github.com/UglyToad/PdfPig/pull/1280/files?w=1


EdwardBlair:
See #1281

Issue #980, opened 2025-01-24
Reading text from files with a lot graphics


iirorahkonen:
Hi,

I'm using PdfPig through Kernel Memory (https://github.com/microsoft/kernel-memory) and ingesting PDF files for text embeddings. Only the text in the documents is relevant to my use cases, and I don't need the graphics. As I understand, I could technically read only the text from PDF files, but it seems that PdfPig doesn't support that.

The issue I face is a huge memory usage, about 4-6 gigs for a 15-megabyte file text and graphics. This is probably mostly caused by the huge list of operations - see the screenshot. 

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/996f8bc9-031d-4959-ad8b-de3e00cb9a1f" />

Does anyone have any insight on how I could get around the issue of the massive memory usage, skip parsing the graphics, or any information related to my assumptions of not needing the graphics?

BobLd:
@iirorahkonen do you have the possibility to share the pdf document in question? I'd like to look into it as I'm currently trying to lower memory usage of PdfPig

You can look into the below on how to only process text command and avoid the rest (both file required):
- https://github.com/CalyPdf/Caly/blob/master/Caly.Pdf/PageFactories/TextLayerFactory.cs
- https://github.com/CalyPdf/Caly/blob/master/Caly.Pdf/TextLayer/TextLayerStreamProcessor.cs

iirorahkonen:
Unfortunately, I cannot share the file as it's proprietary information. It is an A1 size export from Autocad. I can try to find a similar one if that helps.

Thank you for the tip on the page factories! I'll also make sure to test it out. I have already implemented a CustomPdfDecoder in Kernel Memory based on that example for debugging. I'll make sure to share any code for the implementation.

BobLd:
@iirorahkonen no problem with not sharing. As it's from autocad, I expect a lot of path objects indeed. Have a look at what I did in Caly to only extract the text (links I provided before) - let me know if you need help there. I think you'll have good results. Don't hesitate to shares benchmarks too!

EliotJones:
I'm thinking we should expose a `Capabilities` enum flag from parsing options. This would allow skipping slow performance areas when not needed by the client. These capabilities would be things like:

```
[Flags]
public enum PdfParsingCapabilities
{
    Text = 1 << 1,
    ExactGlyphBoundingBoxes = 1 << 2,
    Paths = 1 << 3,
    Images = 1 << 4,
[...]

PR #1281, opened 2026-04-16
Add lazy stream loading to reduce memory for large PDFs

EdwardBlair:
Adds `ParsingOptions.LazyLoading` (off by default) which defers reading stream bytes until they're actually needed. With this on, `page.GetImages()` gives you width, height, bounding box etc without loading the image data into memory - the bytes only get read if you actually access them.

For the file path overload this means using a FileStream instead of File.ReadAllBytes so the data stays on disk. Should help with #1171 where big images cause OOM on cheap nodes.

This could also work as a foundation for #980 - the lazy stream infrastructure means a future capabilities enum could skip processing images/paths entirely and the byte data would never leave disk.


BobLd:
@EdwardBlair thanks for the PR - I need to take time to review it as it touches many areas

src/UglyToad.PdfPig/Parser/PdfDocumentFactory.cs
```diff
@@ -40,6 +40,14 @@ public static PdfDocument Open(string filename, ParsingOptions? options = null)
                 throw new InvalidOperationException("No file exists at: " + filename);
             }
 
+            if (options?.LazyLoading == true)
+            {
+                var stream = new StreamInputBytes(
+                    File.Open(filename, FileMode.Open, FileAccess.Read, FileShare.Read),
+                    true);
+                return Open(stream, options);
+            }
```
cremor, line 49: This requires users of the library to correctly dispose the document. Please add comments to make this very clear, e.g. on `ParsingOptions.LazyLoading`.

See [this commit](https://github.com/UglyToad/PdfPig/pull/1293/changes/18005444d11556719f576b73b40b47b9f69733cd) from a different PR for the full reasoning.

BobLd: agreed with @cremor I think the solution is to not change the code here. Let's leave `public static PdfDocument Open(string filename, ParsingOptions? options = null)` untouched.

There's was debate on how this should open the document, I'm even willing to deprecate this method to only let `byte[]` or `Stream`

EdwardBlair: So if I understand, leave Open(string filename…) as it is (eager) and leave lazy accessible through stream overload?

PR #1283, opened 2026-04-17
Support netstandard2.1

PsykerUdot:


BobLd:
@PsykerUdot can you explain the need to add `netstandard2.1`, especially on top of `netstandard2.0`?

PsykerUdot:
> @PsykerUdot can you explain the need to add `netstandard2.1`, especially on top of `netstandard2.0`?

Unity (www.unity.com) supports only netstandard2.1. Using libraries built against netstandard2.0 can cause dependencies conflict (as Unity itself provides system libraries of certain versions).

BobLd (approved):

cremor:
I don't need .NET Standard 2.1 support (and I don't understand how it would make a difference at runtime), but the way it was done here is not enough to really enable it.
- UglyToad.PdfPig.Package.csproj wasn't updated, so the NuGet package doesn't contain the new target.
- There are many `Condition` attributes in various csproj files that check for `netstandard2.0`. I assume those should have been updated to also check for `netstandard2.1`.

PR #1292, opened 2026-05-12
Xref table expose

vafle228:
## Description

This PR addresses a long-standing limitation for developers needing to implement **incremental PDF updates** (e.g., for digital signatures, form filling, or appending objects without full re‑saving). Currently, the `PdfDocument.Structure` exposes limited low‑level access, because the `CrossReferenceTable` (old property) is missing, and the core token writer lacks support for writing a custom xref table with `/Prev` trailer.

The changes in this PR restore and extend the low-level API to fully support reading and writing **both normal xref tables and xref streams**, as well as working with trailer dictionaries. This makes it possible to:
- Access the raw `CrossReferenceTablePart` entries (object number, offset, generation, type).
- Read/write the trailer dictionary (including `/Prev`, `/Size`, `/Root`, `/Info`).
- Append new objects while keeping the original xref section intact.
- Implement full incremental PDF update workflows.

## Changes Made
1. **`CrossReferenceTable`** is now fully exposed via `PdfDocument.Structure.CrossReferenceTable`
   (previously it was removed). It provides:
   - `ObjectOffsets` – mapping from `IndirectReference` to `XrefLocation`.
   - `Parts` - single instance of xref table
2. **`TrailerDictionary`** now exposes `/Prev`, `/Size`, `/Root`, `/Info` (and all raw dictionary entries).
3. **`TokenWriter`** has been extended with a new method:
   ```csharp
   public void WriteCrossReferenceTable(
       IReadOnlyDictionary<IndirectReference, long> objectOffsets,
       IndirectReference catalogToken,
       Stream outputStream,
       IndirectReference? documentInformationReference,
       long? prevTableLocation);   // 👈 new parameter for incremental save
4. Added  `AdvancedMerge.cs` example of low level tools usage


src/UglyToad.PdfPig.Tokens/DictionaryToken.cs
```diff
@@ -1,171 +1,180 @@
...
+        }
+        
+        /// <summary>
+        /// Creates empty DictionaryToken instance
+        /// </summary>
+        /// <returns>A new empty <see cref="DictionaryToken"/></returns>
+        public static DictionaryToken Empty()
+        {
+            return new DictionaryToken(new Dictionary<string, IToken>());
+        }
```
BobLd: why not `public static readonly DictionaryToken Empty ...`

src/UglyToad.PdfPig/CrossReference/CrossReferenceTablePart.cs
```diff
@@ -21,13 +23,13 @@
     /// giving the number of bytes from the beginning of the file to the beginning of the
     /// object.
     /// </remarks>
-    internal class CrossReferenceTablePart
+    public class CrossReferenceTablePart
     {
-        public IReadOnlyDictionary<IndirectReference, long> ObjectOffsets { get; }
+        public IReadOnlyDictionary<IndirectReference, XrefLocation> ObjectOffsets { get; }
 
         public long Offset { get; private set; }
```
BobLd: no need for `private set;`, read only

src/UglyToad.PdfPig/CrossReference/CrossReferenceTablePart.cs
```diff
@@ -21,13 +23,13 @@
...
 
-        public long Previous { get; }
+        public long? Previous { get; }
 
         public DictionaryToken Dictionary { get; private set; }
```
BobLd: no need for `private set;`, read only

src/UglyToad.PdfPig/CrossReference/CrossReferenceTable.cs
```diff
@@ -13,49 +15,67 @@
     /// </summary>
     public class CrossReferenceTable
     {
-        private readonly Dictionary<IndirectReference, long> objectOffsets;
+        private readonly List<CrossReferenceTablePart> parts;
```
BobLd: Change to `CrossReferenceTablePart[]`, no need for a list from what I see

src/UglyToad.PdfPig/CrossReference/CrossReferenceTable.cs
```diff
@@ -13,49 +15,67 @@
...
+            
+            Type = parts.FirstOrDefault()?.Type ?? CrossReferenceType.Table;
             Trailer = trailer ?? throw new ArgumentNullException(nameof(trailer));
-            CrossReferenceOffsets = crossReferenceOffsets ?? throw new ArgumentNullException(nameof(crossReferenceOffsets));
-
-            var result = new Dictionary<IndirectReference, long>(capacity: objectOffsets.Count);
-            foreach (var objectOffset in objectOffsets)
+            this.objectOffsets = objectOffsets.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+            
+            var result = new List<CrossReferenceOffset>();
```
BobLd: Use a `CrossReferenceOffset[]` instead

src/UglyToad.PdfPig/CrossReference/CrossReferenceTable.cs
```diff
@@ -13,49 +15,67 @@
...
+            for (var i = 1; i < parts.Count; i++)
             {
-                result[objectOffset.Key] = objectOffset.Value;
+                result.Add(new CrossReferenceOffset(parts[i].Offset, parts[i - 1].Previous));
```
BobLd: can you double check that it should not be the below instead:
```csharp
foreach (var part in parts)
{
      result.Add(new CrossReferenceOffset(part.Offset, part.Previous));
}
```

BobLd (changes requested):
you can also un-comment the asserts in `PdfMergerTests`

BobLd:
I have review your PR, let me know if any question, great job!

BobLd:
Also, review `AdvancedMerge` and make sure it's referenced into `Program.cs`

src/UglyToad.PdfPig/CrossReference/CrossReferenceTable.cs
vafle228: Thank for that

vafle228:
@BobLd Thank you for the review! I’ve applied all the suggested changes:
- Replaced `Empty()` method with `static readonly Empty` field.
- Removed unnecessary `private set;` from read‑only properties.
- Changed `List<CrossReferenceTablePart>` to array.
- Simplified `CrossReferenceOffset` creation logic using `foreach`.
- Verified `AdvancedMerge` is now referenced in `Program.cs` for the samples.

Could you please take another look when you have time? Thanks again!

examples/Program.cs
```diff
@@ -47,12 +47,24 @@ public static void Main()
...
+                } ,
+                {9,
+                    ("Advanced pdf merge, using low level pdf tools, like trailer dictionary and xref table",
+                    () =>
+                    {
+                        using var input2 = File.Open(Path.Combine(filesDirectory, "EmptyPdf.pdf"), FileMode.Open);
+                        using var output = new FileStream(Path.Combine(filesDirectory, "AdvancedMergeResult.pdf"), FileMode.Create);
+                        using var input = File.Open(Path.Combine(filesDirectory, "Single Page Simple - from google drive.pdf"), FileMode.Open);
+
+                        input2.CopyToAsync(output);
```
BobLd: `input2.CopyToAsync(output);` is not awaited, `input2.CopyTo(output);` should be enough

src/UglyToad.PdfPig/Writer/ITokenWriter.cs
```diff
@@ -33,10 +33,13 @@ public interface ITokenWriter
         /// <param name="catalogToken">The object representing the catalog dictionary which is referenced from the trailer dictionary.</param>
         /// <param name="outputStream">The output stream to write to.</param>
         /// <param name="documentInformationReference">The object reference for the document information dictionary if present.</param>
+        /// /// <param name="prevTableLocation">The offset to the previous xref table if present</param>
```
BobLd: duplicate `///`

examples/Program.cs
```diff
@@ -47,12 +47,24 @@ public static void Main()
...
                     () => OpenDocumentAndExtractWords.Run(Path.Combine(filesDirectory, "OPEN.RABBIT.ENGLISH.LOP.pdf")))
-                } 
-        };
+                } ,
+                {9,
+                    ("Advanced pdf merge, using low level pdf tools, like trailer dictionary and xref table",
+                    () =>
+                    {
+                        using var input2 = File.Open(Path.Combine(filesDirectory, "EmptyPdf.pdf"), FileMode.Open);
+                        using var output = new FileStream(Path.Combine(filesDirectory, "AdvancedMergeResult.pdf"), FileMode.Create);
```
BobLd: let's not write the output document to the integration tests folder. `new FileStream("AdvancedMergeResult.pdf")` should be enough

BobLd (changes requested):
Can you also review the `AdvanceMerge` example? It produces a blank page. Can you also add this example as an integration test, if possible?

BobLd:
@vafle228 I've added some more comments, let me know if any question. Can you also add an integration test based on `AdvanceMerge`? Thanks again

vafle228:
@BobLd Thank you for the new comments! My bad with the `AdvancedMerge.cs` example. I forgot to link resources, so readers couldn't render the content properly. I fixed this in the last two commits. Also I added the integration test, as you asked for. Let me know if there are any issues left.

BobLd (approved):

BobLd:
note to self: 
> Trailer-merging semantics changed silently. The deleted builder special-cased `/Size` so a linearized PDF's first trailer wins. The new FirstPassParser walks all parts and prefers the dictionary that has `/Root`. This is reasonable but is a behavior change, worth a targeted test on a linearized file to confirm `/Size` hasn't regressed.

PR #1293, opened 2026-05-14
Support opening PDFs larger than 2 GiB

### LINKED ISSUE #1259: How to support files larger than 2GB ?
File.radAllbytes() does not support 2GB+files

Image

OrganizationUsername:
Fixes `PdfDocument.Open(string)` for PDFs larger than 2 GiB by opening the file as a seekable stream instead of reading the whole file into a byte array with `File.ReadAllBytes`.

Also widens `FileHeaderOffset` from `int` to `long` so header offsets are not narrowed during parsing.

Adds regression coverage using a sparse PDF with a logical size over 2 GiB, verifying that PdfPig can open it, read page 200, and extract text. A skipped local repro test is also included for generating and reading a real >2 GiB PDF.

Fixes #1259.


OrganizationUsername:
Happy to drop the skipped real-file repro if it makes the PR easier to review. I included it because it’s pretty close to the kind of PDF I actually have to deal with at work.

BobLd:
@OrganizationUsername yes, do you mind removing these optional test? the rest looks good to me. Thanks for that!

BobLd:
@OrganizationUsername  can you also drop `LocalTests`?

cremor:
> Fixes `PdfDocument.Open(string)` for PDFs larger than 2 GiB by opening the file as a seekable stream instead of reading the whole file into a byte array with `File.ReadAllBytes`.

Seems like this is no longer true. Please update the description of this PR.

BobLd (approved):
approving as-is and I'll remove the dead test

EDIT: Nothing to remove - misread github diffs

PR #1341, opened 2026-06-22
Fix/representation of more letters with the same position

gherards99:
## Summary

Some PDF producers draw the same text more than once at the exact same location, for example overprinting, or faux-bold achieved by "double-striking" the same glyph. A PDF viewer renders the copies on top of one another, so the text is only *seen* once. However, `DefaultWordExtractor` reported every glyph, which duplicated characters in the extracted text (a visible `12A` could come out as `1122A`).

This PR makes `DefaultWordExtractor.GetWords` skip a glyph when it shares the same baseline coordinate as the previous one, so overlapping copies are collapsed into a single character.

## What changed

- Added a `SamePosition(Letter a, Letter b)` check that compares the baseline location of two letters within a small tolerance (`SamePositionTolerance = 1e-3`, in PDF user-space units).
- During word extraction, when the current letter occupies the same position as the previous letter, it is dropped.

If you need the file I used for testing, please let me know and I can share it with you privately, as it is confidential and cannot be shared publicly.

If you think this could be handled in a better way, please feel free to let me know.

Thanks in advance for your time.


BobLd:
@gherards99 thanks a lot for the PR. The faux-bold is currently handled here: https://github.com/UglyToad/PdfPig/blob/master/src/UglyToad.PdfPig.DocumentLayoutAnalysis/DuplicateOverlappingTextProcessor.cs

I agree that this should be included in `DefaultWordExtractor`, do you think you could reuse that?

gherards99:
Hi @BobLd,

Thank you for the quick response, I really appreciate it.

I'm taking a look at it now and will let you know how it goes.

Thanks for pointing me in the right direction, and sorry for overlooking DuplicateOverlappingTextProcessor.

gherards99:
Hi @BobLd,

For now, the easiest solution on my side is to call DuplicateOverlappingTextProcessor.Get(letters) before GetWords and pass it the cleaned set of letters, that should already do the job.

```
var wordExtractor = new NearestNeighbourWordExtractor(wordExtractorOptions);
var extractedLetters = DuplicateOverlappingTextProcessor.Get(letters);
var words = wordExtractor.GetWords(extractedLetters);
```
As for the PR, when I have some spare time I'd like to take the following approach:

-Move the DuplicateOverlappingTextProcessor class into the core project, so it's available in more places than it is today.
-Keep a class with the same name in its current location that simply forwards to the moved one, so existing calls don't break.
-Then it could be nice to add an option to choose whether the letters should be cleaned before building the words, or                     
  whether the raw list should be used as-is.

If you think this is a good direction, I'll try to implement it; otherwise, we can discuss it first.

Thanks!

BobLd:
@gherards99 I agree with your overall point.
> -Move the DuplicateOverlappingTextProcessor class into the core project, so it's available in more places than it is today.

The question about "fake-bold" comes up regularly, which makes me think it is currently not located in the right place.

> -Keep a class with the same name in its current location that simply forwards to the moved one, so existing calls don't break.

Fine by me, but the old location should be marked as `[Obsolete]` (will be removed in later versions).

> -Then it could be nice to add an option to choose whether the letters should be cleaned before building the words, or
whether the raw list should be used as-is.

Also make sense, but not sure about the best approach

PR #1359, opened 2026-07-02
Expose annotation appearance streams and appearance state as public

jeske:
This is necessary for my PDF Renderer to access the Appearance Stream parsing inside PdfPig..

I can do it without this, but it means duplicating the parsing that is already there. 

Here is a screenshot of PDF Highlight Annotations Rendering. 

>  Note: the text looks washed out because we are temporaily emulating multiply blend mode with alpha, but that's an issue inside our renderer not related to PdfPig.

image

### DIFF
--- a/src/UglyToad.PdfPig/Annotations/Annotation.cs
+++ b/src/UglyToad.PdfPig/Annotations/Annotation.cs
@@ -84,6 +84,26 @@
         public bool HasDownAppearance => downAppearanceStream != null;
 
         /// <summary>
+        /// The normal appearance stream for this annotation, if any. Each appearance is a Form XObject.
+        /// </summary>
+        public AppearanceStream? NormalAppearance => normalAppearanceStream;
+
+        /// <summary>
+        /// The roll-over appearance stream (shown on hover), if any.
+        /// </summary>
+        public AppearanceStream? RollOverAppearance => rollOverAppearanceStream;
+
+        /// <summary>
+        /// The down appearance stream (shown on click), if any.
+        /// </summary>
+        public AppearanceStream? DownAppearance => downAppearanceStream;
+
+        /// <summary>
+        /// The current appearance state name, if any.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


BobLd:
@jeske ill have a look shortly but do you mind having a look at https://github.com/BobLd/PdfPig.Rendering.Skia

This is my own pdf rendered (it renders annotations too). Not saying this is the only way to do a rendered but the approach there expose everything you'd need

Edit: it is using Skia, but you could use any another graphic engine

jeske:
> @jeske ill have a look shortly but do you mind having a look at https://github.com/BobLd/PdfPig.Rendering.Skia
> 
> This is my own pdf rendered (it renders annotations too). Not saying this is the only way to do a rendered but the approach there expose everything you'd need
> 
> Edit: it is using Skia, but you could use any another graphic engine

Can you explain why you are suggesting this?

Of course i can reparse annotations myself, as i said in my PR... "I can do it without this, but it means duplicating the parsing that is already there."

PdfPig has Annotations parsing classes that as far as i can see are not used internally but also not exposed.

what are they for if not to be exposed and used? I figured it would be good to expose and use them.

If this patch isnt wanted, i can go the other route, but down this route it seems like PdfPig would delete annotations parsing that is not used and not exposed?

Or did i miss something? Is there a way to get at the data of the internal annotations classes i missed?

Humbly trying to help.

BobLd:
@jeske no worries at all. Will check your changes shortly. I believe they are indeed needed

BobLd (approved):

jeske:
@BobLd  I see now what you meant... and the BaseStreamProcessor<T> method... I'm going to refactor to use this now more like your Skia renderer. Thanks!

BobLd:
@jeske if you see any changes in PdfPig that would simplify the logic for rendering, do not hesitate. 

Also, if you have any question about the skia rendered, do not hesitate to open a discussion there

PR #1367, opened 2026-07-16
Expose font dictionary indirect reference on FontDetails

jeske:
Documents can embed multiple font objects sharing one BaseFont name (e.g. two different subsets of a typeface embedded without unique subset prefixes). Consumers resolving fonts by Letter.FontName alone cannot tell them apart, causing glyph-level rendering corruption when the subsets have complementary glyph coverage.

- FontDetails.FontDictionaryReference: the indirect reference of the font dictionary the font was loaded from (null for direct/synthetic fonts); propagated through AsBold()/WithName()
- ResourceStore stamps the reference at both font-load sites (LoadFontDictionary and GetFontDirectly)

This gives Letter.FontDetails a stable in-document font identity without changing the IFont interface or any rendering behavior.

Note: I'm using BaseStreamProcessor now, but unlike the Skia renderer which relies on PdfPig's fallback per-glyph shape drawing, my SafePDF resolves fonts using the font-definitions so I can install fonts into my rasterized and hinted Font Glyph Atlas for faster screen drawing, and into D2D as a font for smaller print spool streams. If you have a better idea of how to do the above, let me know.

### DIFF
--- a/src/UglyToad.PdfPig/Content/ResourceStore.cs
+++ b/src/UglyToad.PdfPig/Content/ResourceStore.cs
@@ -232,7 +232,12 @@
 
                     try
                     {
-                        loadedFonts[reference] = fontFactory.Get(fontObject);
+                        var loadedFont = fontFactory.Get(fontObject);
+                        // Stamp the font dictionary's indirect reference so consumers can
+                        // distinguish same-named fonts (e.g. two subsets of one typeface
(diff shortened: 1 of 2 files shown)


BobLd (approved):

BobLd:
@jeske  thanks a lot for that.

I'll have a look at my own skia code too.

PR #1335, opened 2026-06-20
Add ICC profile support

BobLd:
Adds optional ICC profile support to color handling, through a new `ParsingOptions.IccProfileService` and 3 new public interfaces (`IIccProfileService`, `IIccProfile`, `IIccTransform`) that resolve raw profile bytes into reusable, intent-aware, sRGB-producing transforms.

Rendering intent is now passed through the color pipeline via overloads on `ColorSpaceDetails`, with the param-less versions delegating to the PDF default (RelativeColorimetric); the graphics-state intent is propagated from `ColorSpaceContext` and image `RenderingIntent` flows through `ColorSpaceDetailsByteConverter` and `PngFromPdfImageFactory`.

`ICCBasedColorSpaceDetails` now resolves a profile when a service is configured, and otherwise falls back silently to its alternate color space. `ResourceStore` gained default color space substitution (DefaultGray/DefaultRGB/DefaultCMYK) and exposes the configured IIccProfileService.

A shared `DenormalizeComponents` helper centralizes mapping normalized [0,1] components to a space's native range (identity by default, overridden for  Lab).

No actual implementation of ICC processing is made available directly in PdfPig. This PR should be seen in combination with https://github.com/BobLd/PdfPig.Rendering.Skia/pull/103 This uses [Wacton.Unicolour](https://github.com/waacton/Unicolour) (cc @waacton for visibility) as ICC profile processor. `UnicolourIccProfile` is for the moment sitting with [PdfPig.Rendering.Skia](https://github.com/BobLd/PdfPig.Rendering.Skia) but will be split in its own NuGet package shortly.

PR #1369, opened 2026-07-18
BUGFIX: Decode Indexed colour-table bytes to the base space's component ranges

jeske:
## Problem: Indexed color decode bug..

There is an indexed color decode bug breaking the "Lab" (lower left corner) square in this indexed color test.

> NOTE: some *other* colors in this test vary slightly because Adobe is applying an ICC print color adjustment, and we are not. AFAIK these are not bugs.. just the lower left square.

(top acrobat, bottom PdfPig colorspace)
image

Sample file: 
[pdfium_2_color_indexed.pdf](https://github.com/user-attachments/files/30144825/pdfium_2_color_indexed.pdf)

## Explanation of the fix...

ISO 32000-2 8.6.6.3: colour table data is interpreted as component values in the base colour space. IndexedColorSpaceDetails unconditionally scaled table bytes to [0, 1], which is only correct for spaces whose components span [0, 1]. Lab expects L* in [0, 100] and a*/b* in the /Range entry (default [-100, 100]) - an Indexed-over-Lab lookup therefore produced L* <= 1.0, rendering near-black for any table entry (e.g. white table entries render black).

Scale each component to the base space's range: Lab per-component (L* x100, a*/b* lerped over /Range); all other supported base spaces keep the [0, 1] scaling.

### DIFF
--- a/src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
+++ b/src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
@@ -403,15 +403,7 @@
         internal override double[] Process(params double[] values)
         {
             var csBytes = UnwrapIndexedColorSpaceBytes([ClampColorIndex(values[0])]);
-
-            var scaledCsBytes = new double[csBytes.Length];
-
-            for (int i = 0; i < csBytes.Length; i++)
-            {
-                scaledCsBytes[i] = csBytes[i] / 255.0;
-            }
-
-            return BaseColorSpace.Process(scaledCsBytes);
+            return BaseColorSpace.Process(ScaleIndexedComponents(csBytes));
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


BobLd:
@jeske thanks a lot for that! That's odd the bug is still there because I believe I fixed it some time ago. How is UglyToad.PdfPig.Rendering.Skia rendering this document? I'm away at the moment, and will check when I'm back if you can't.

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
```diff
@@ -425,16 +417,43 @@ public override IColor GetColor(params double[] values)
...
-                var scaledCsBytes = new double[csBytes.Length];
+        /// <summary>
+        /// Decode colour-table bytes to the base colour space's component ranges.
+        /// ISO 32000-2 (PDF 2.0) 8.6.6.3: the colour table data is interpreted as
+        /// component values in the base space, i.e. each byte decodes to
+        /// min + (byte / 255) × (max − min) of that component's range. Device and
+        /// most CIE spaces use [0, 1]; Lab uses L* ∈ [0, 100] and a*/b* from the
+        /// /Range entry — feeding Lab a [0, 1] L* renders near-black.
+        /// </summary>
+        private double[] ScaleIndexedComponents(Span<byte> csBytes)
```
BobLd, line 432: Can you update the method to do the scaling in place?

BobLd (changes requested):

BobLd:
thanks again @jeske , I have put one comment. If you are not available, I'm happy to merge it as is and do the changes.

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
jeske: I would be happy to, but i dont understand what you are asking for. This converts a PDF byte-stream index color table into the double Lightness/a-green-red/b-blue-yellow values. I can't "edit the byte span in place" because the output data is not of the same type.

Did you mean that you want it to accept a double-array or double-span buffer that has to be preallocated to fit the output?

BobLd: My bad, I should have had more coffee - you can ignore my remark

BobLd (approved):

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
jeske: Do you want this to be `void ScaleIndexedComponents(Span<double>)` where the caller converts bytes to double and then we scale it? 

While looking at this, it appear GetRgb may have the same Lab scaling bug that would caused indexed-Lab images to rasterize near black.

```csharp 
case ColorSpace.DeviceRGB:
case ColorSpace.CalRGB:
case ColorSpace.Lab:          // <-- Lab lumped in
    buffer[j] = colorTable[index * 3 + j] / 255.0;   // only [0,1] scaling!
```

If this was `void ScaleIndexedComponents(Span<double>)` then it could be applied to the output doubles in Process, GetColor and GetRgb... is that what you mean?

jeske: actually, im looking at another approch to fix the fact that this color space conversion is located in three places, which is why even my fix didn't fix all three cases because i didn't notice the others. i'll have a patch in a little bit... and you can review it

jeske:
Ohh, i see you meged this shortly after my comment today... I have a different version of this now... i'lll make a new branch and see how your changes landed first

src/UglyToad.PdfPig/Graphics/Colors/ColorSpaceDetails.cs
BobLd: @jeske I misread your code and was thinking the input span was `Span<double>`, so you are correct and in-place is not possible - at least with the current methods. Using a buffer could be the way forward.

> actually, im looking at another approch to fix the fact that this color space conversion is located in three places, which is why even my fix didn't fix all three cases because i didn't notice the others. i'll have a patch in a little bit... and you can review it

That's perfect, thanks a lot for your help here - really appreciate it. I believe I touched upon this issue in my draft PR on ICC profile [here](https://github.com/UglyToad/PdfPig/pull/1335). This PR will not be merged as is as the scope expended too much. I started splitting it in smaller pieces and your work on this will be greatly helpful (That being said, you can 100% ignore my draft PR as part as your patch)

BobLd:
@jeske no problem at all, I saw your comment shortly after merging. Let me know if I can help

PR #1375, opened 2026-08-05
Move Indexed table-byte decoding onto the base colour space (in place)

jeske:
This is a continuation of the Lab color space fix... 

- Fixes GetRgb Lab indexed color handling - previously table bytes were decoded to [0,1] instead of Lab native ranges L [0,100], a/b per /Range, this caused black Lab colors on the GetRgb path (Same bug as my previous fix in Process / GetColor) - now fixed
- Replaces ScaleIndexedComponents with internal virtual void DecodeRawComponents(ReadOnlySpan<byte> raw, Span<double> destination), letting the stackalloc GetRgb path share the same decode; LabColorSpaceDetails overrides it, so the Lab rule lives in exactly one place and the per-base-type switch in GetRgb goes away
- GetRgb now rounds and clamps the colour index per [PDF 8.6.6.3 Indexed Colour Spaces](https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf#page=164) instead of casting to byte (which truncated fractions and wrapped out-of-range values)
- Adds tests covering all three decode entry points (GetColor, GetRgb, Process).
- NOTE: this removed the default: case of an unknown color-space-id, though the result will still be an InvalidOperationException on Unsupported/unknown/malformed color space
- NOTE: a short/malformed color table would throw on old code, and will throw on this code too, hardening could make this more robust if that's desired (I wrap a guard for this around calls to color.ToRGBValues() in my code)

### DIFF
--- a/src/UglyToad.PdfPig.Tests/Graphics/Colors/IndexedLabColorSpaceDetailsTests.cs
+++ b/src/UglyToad.PdfPig.Tests/Graphics/Colors/IndexedLabColorSpaceDetailsTests.cs
@@ -1,0 +1,103 @@
+namespace UglyToad.PdfPig.Tests.Graphics.Colors
+{
+    using PdfPig.Graphics.Colors;
+
+    /// <summary>
+    /// Indexed colour-table bytes must decode to the base colour space's native component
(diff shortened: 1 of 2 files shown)


BobLd:
Looks good, thx a lot @jeske

BobLd (approved):

PR #1422, opened 2026-08-26
Added possibility to set additional directories to search fonts

adelobosko:
• fixed typo in WindowsSystemFontLister (PSFonts dir was not included, fonts dir was used twice instead)

• added possibility to set additional directories to search fonts

useful for strict environments without preinstalled fonts or without access to them
related to https://github.com/BobLd/PdfPig.Rendering.Skia/issues/145


src/UglyToad.PdfPig.Fonts/SystemFonts/AndroidSystemFontLister.cs
```diff
@@ -6,13 +6,28 @@
...
+        }
+
+        private IEnumerable<SystemFontRecord> GetAllFonts(IEnumerable<string>? additionalDirectories)
         {
             var directories = new List<string>
             {
                 "/system/fonts",
             };
 
+            if (additionalDirectories != null)
```
BobLd: `is not null` instead of `!=` (same for all other occurrences)

src/UglyToad.PdfPig.Fonts/SystemFonts/AndroidSystemFontLister.cs
```diff
@@ -6,13 +6,28 @@
 
     internal sealed class AndroidSystemFontLister : ISystemFontLister
     {
-        public IEnumerable<SystemFontRecord> GetAllFonts()
+        IEnumerable<SystemFontRecord> ISystemFontLister.GetAllFonts()
+        {
+            return GetAllFonts(null);
+        }
+
+        IEnumerable<SystemFontRecord> ISystemFontLister.GetAllFonts(IEnumerable<string> additionalDirectories)
```
BobLd: Unless I missed something, the method is not needed, remove `ISystemFontLister.GetAllFonts(...` and make the below method public (with the nullable change in the interface):
```csharp
private IEnumerable<SystemFontRecord> GetAllFonts(IEnumerable<string>? additionalDirectories)

becomes

public IEnumerable<SystemFontRecord> GetAllFonts(IEnumerable<string>? additionalDirectories)

```

src/UglyToad.PdfPig.Fonts/SystemFonts/ISystemFontLister.cs
```diff
@@ -5,5 +5,6 @@
     internal interface ISystemFontLister
     {
         IEnumerable<SystemFontRecord> GetAllFonts();
+        IEnumerable<SystemFontRecord> GetAllFonts(IEnumerable<string> additonalFontSearchPaths);
```
BobLd: I think you can make `IEnumerable<string>` nullable to match my other comment

src/UglyToad.PdfPig.Fonts/SystemFonts/SystemFontFinder.cs
```diff
@@ -26,6 +26,11 @@ public sealed class SystemFontFinder : ISystemFontFinder
     /// </summary>
     private static readonly Lazy<IReadOnlyDictionary<char, SystemFontRecord[]>> FontsByFirstChar;
 
+    /// <summary>
+    /// Additional directories where to search fonts
+    /// </summary>
+    public static readonly ConcurrentBag<string> AdditionalFontSearchDirectories = new ConcurrentBag<string>();
```
BobLd: in term of public api, can you make that private and add a method to add a directory. Something along these lines:
```csharp
    /// <summary>
    /// Add a directory to the list of additional directories where to search for fonts.
    /// </summary>
    /// <param name="directory"></param>
    public static void AddFontSearchDirectory(string directory)
    {
        if (!string.IsNullOrWhiteSpace(directory))
        {
            AdditionalFontSearchDirectories.Add(directory);
        }
    }

    /// <summary>
    /// Additional directories where to search fonts
    /// </summary>
    private static readonly ConcurrentBag<string> AdditionalFontSearchDirectories = new ConcurrentBag<string>();
```

BobLd (changes requested):

BobLd:
@adelobosko thanks a lot for the PR, I think this is going to help many users.

I have added comments - let me know if any questions

The feedback in `AndroidSystemFontLister.cs` is applicable to all other `ISystemFontLister` implementations

BobLd:
@adelobosko I did a 2nd pass on the code, have you successfully added other fonts in your use case?

`AdditionalFontSearchDirectories` is used inside `Lazy<>`, so once the lazy value is actually created, adding new font path won't do anything, and this happens silently.

We should aim at throw an exception if a path is added too late.

adelobosko:
@BobLd Yes, I was able to successfully add other fonts in my use case.

BobLd:
@adelobosko thanks for updating the PR.

Regarding my last comment, your `FontSearchDirectoryLock` does not do anything for the moment (and it's gating a concurrent bag).

Can you review your changes?

As a side note: I agree with the approach being static (even if not a huge fan, but I don't think there's an alternative), it just need to be robust.

adelobosko:
> Regarding my last comment, your `FontSearchDirectoryLock` does not do anything for the moment (and it's gating a concurrent bag).

`FontSearchDirectoryLock` preventing duplicates, maybe better to use HashSet<string> + lock instead of Concurrent bag

BobLd:
@adelobosko agreed on the hashset, yes. My comment above was related to the use of `AvailableFonts.IsValueCreated`. It is not enough as a gate, `FontSearchDirectoryLock` should also be lock inside the lazy creation method.

Will merge as-is and push changes for the lock, and hashset.

adelobosko:
@BobLd done, changed to HashSet + locks

BobLd:
@adelobosko thx for the updates. Can you let me know when you are done? I'll run the tests and merge

adelobosko:
@BobLd It's done

BobLd (approved):

Issue #1144, opened 2025-09-03
Add support for ICC profile - Colors from CMYK  are different in RGB


RodiNikolaev:
Hello,

I'm trying to import a PDF file, but the colors are different.

 This is my code:

``` C#
  var fill = path.FillColor?.ToRGBValues();

  SvgColor? svgFill = fill != null ?
       new SvgColor((byte)(fill.Value.r * 255), (byte)(fill.Value.g * 255), (byte)(fill.Value.b * 255)) : null;
```
Example.

The original hex color: #00973f
The CMYK color in the debugger: CMYK(1,0,1,0)
The result color: #00ff00

<img width="699" height="759" alt="Image" src="https://github.com/user-attachments/assets/046e0409-37cc-4066-8bad-b2bd451d9926" />

BobLd:
@RodiNikolaev can you share the pdf document in question?

waacton:
@BobLd I hope you don't mind me poking around at this, I was just wondering how hard it would be to use Unicolour as the _"implement your own `ToRGBValues()` method"_.

I was able to get close to the expected value by doing:

```c#
var fill = path.FillColor as CMYKColor;

var iccConfig = new IccConfiguration("Coated_Fogra39L_VIGC_300.icc", Intent.RelativeColorimetric);
var config = new Configuration(iccConfig: iccConfig);
var unicolour = new Unicolour(config, new Channels(fill.C, fill.M, fill.Y, fill.K)); // RGB -0.32 0.60 0.28
var inGamut = unicolour.MapToRgbGamut(GamutMap.RgbClipping);                         // RGB 0.00 0.60 0.28
```

CMYK(1, 0, 1, 0) here gets converted to #009847, which is pretty close to the original rendering of #00973F.

The FOGRA39 profile here is just a guess. If you were able to expose the ICC profile bytes related to each path, a user could use Unicolour (or some other ICC conversion code) to get a more precise RGB value - and you might not have to worry about handling ICC conversions (or their performance costs) in PdfPig for the time being.

BobLd:
@waacton apologies for the delay, I forgot to get back to you.

On the contrary, your expertise with ICC profiles is more than welcome.

I need to think how to give PdfPig users the possibility to easily provide ICC library like Unicolour to handle ICC conversion. I have to admit I have left that on the back burner for quite some time now.

For context, in a pdf document, ICC profiles can be present in 2 ways:
- At path, letter, etc. level
- At image level

[...]

Issue #1054, opened 2025-05-30
"The SMask dictionary contains a 'ColorSpace'." exception


BobLd:
The following document throws a "The SMask dictionary contains a 'ColorSpace'." exception, but the images are in fact valid.

As per the Pdf Specifications:  "The image dictionary shall not contain a ColorSpace entry because sample values represent masking properties (1 bit per sample) rather than colours." but because we should not throw an exception as the image can still be used.

TODO for later: we should pass the parsing options and throw an exception only if Lenient parsing is false.

[MOZILLA-11518-0.pdf](https://github.com/user-attachments/files/20516076/MOZILLA-11518-0.pdf)

PR #1424, opened 2026-08-29
Add ICC profile, rendering intent and output intent support

BobLd:
_The code changes were made with AI assistance._

PdfPig can now support ICC profiles for colour management.

`ParsingOptions.IccProfileService` now lets a caller plug in a colour management module (like [Unicolour](https://github.com/waacton/Unicolour)). No ICC profile management is done in PdfPig.

When one is configured, `/ICCBased` colour spaces convert through their embedded profile instead of always deferring to the alternate, and a `/JPXDecode` image that declares no `/ColorSpace` decodes the profile embedded in its own codestream. Output intents are parsed - the document catalog's, and in PDF 2.0 a page's own, which override it - exposed on `IResourceStore` and on the graphics state, and, when the service asks for it, applied: device colours (DeviceGray, DeviceRGB, DeviceCMYK) are converted through the output intent's profile rather than the built-in approximations. Rendering intent is now passed through every colour conversion, and a colour selected from operands keeps them, so a later `ri` reconverts it instead of leaving it pinned to the intent in force when it was selected.

**Without an ICC profile service configured, behaviour is unchanged.**

### Result in image

Page on the left is a page rendered with PdfPig without `IccProfileService`, the page on the right had one based on [Unicolour](https://github.com/waacton/Unicolour) provided:
image
[version4pdf.pdf](https://github.com/user-attachments/files/31146641/version4pdf.pdf)

### How to use
Since an `IccProfileService` is required, a user will either need to create their own implementation, or use one provided by a third party. Once this PR is merged, I will provide a NuGet package implementation.

### Opting out entirely

Without `ParsingOptions.IccProfileService` set, no ICC colour management happens and `/ICCBased` spaces behave as before. With one set but `UseOutputIntent` set to false, `/ICCBased` spaces are colour-managed and device colours are left alone, which is the recommended default: 14.11.5 sets "no expectation" that device colours are converted to the output intent's target, and says such conversion is "undesirable" in some workflows.

---

## New public API

| Member | |
|---|---|
| `ParsingOptions.IccProfileService` | The opt-in. `null` by default. |
| `Graphics.Colors.Icc.IIccProfileService` | What a caller implements. |
| `Graphics.Colors.Icc.IIccProfile` / `IIccTransform` | What the service hands back: a profile, and a transform bound to one rendering intent. |
| `Graphics.Colors.Icc.OutputIntent` / `IccProfileReference` | A parsed `/OutputIntents` entry, and its PDF 2.0 `/DestOutputProfileRef`. |
| `Graphics.Colors.Icc.OutputIntentColorManagement` | Picks the intent to manage through, and converts device colours and device image data through it. |
| `IResourceStore.DocumentOutputIntents`, `GetPageOutputIntents`, `GetPageOutputIntentProfile`, `IccProfileService` | |
| `CurrentGraphicsState.OutputIntentProfile` | The profile device colours in the content being processed are managed through, or `null`. |
| `ICCBasedColorSpaceDetails.IccProfile` | The resolved profile, or `null` when there is no service or it declined the bytes. |
| `ColorSpaceDetails.RenderingIntentAffectsOutput` | Whether this space can answer differently per intent. |
| `ColorSpaceDetails.GetColor` / `GetRgb` / `GetInitializeColor` overloads taking a `RenderingIntent` | The old signatures stay, as overloads that pass `RelativeColorimetric` (the PDF default). |

---

### 1. `BaseStreamProcessor<T>` constructor gained `outputIntentProfile`

Affects custom stream processors.

```csharp
protected MyStreamProcessor(
    int pageNumber,
    IResourceStore resourceStore,
    IPdfTokenScanner pdfScanner,
    IPageContentParser pageContentParser,
    ILookupFilterProvider filterProvider,
    CropBox cropBox,
    UserSpaceUnit userSpaceUnit,
    PageRotationDegrees rotation,
    in TransformationMatrix initialMatrix,
+   IIccProfile? outputIntentProfile,
    ParsingOptions parsingOptions)
    : base(pageNumber, resourceStore, pdfScanner, pageContentParser, filterProvider,
           cropBox, userSpaceUnit, rotation, initialMatrix,
+          outputIntentProfile,
           parsingOptions)
```

It seeds `CurrentGraphicsState.OutputIntentProfile` for the initial state. Pass `null` for no colour management of device colours, or ask the resource store for the page's:

```csharp
ResourceStore.GetPageOutputIntentProfile(pageDictionary)
```

which answers `null` unless there is a service, it opted in with `UseOutputIntent`, and an output intent with a usable `/DestOutputProfile` is in effect. It takes the page dictionary so a page-level `/OutputIntents` (PDF 2.0, Table 31) is picked up; pass `null` and the document catalog's intents are used.

---

### 2. `XObjectFactory.ReadImage` gained `ParsingOptions`

```csharp
- XObjectFactory.ReadImage(record, scanner, filterProvider, resourceStore);
+ XObjectFactory.ReadImage(record, scanner, filterProvider, resourceStore, options);
```

Inside a `BaseStreamProcessor` subclass, `ParsingOptions` is already available as a property, so the call becomes `..., resourceStore, ParsingOptions)`.

The options carry `IccProfileService`, which is what lets a JPEG 2000 image decode the ICC profile embedded in its own codestream. They also carry `UseLenientParsing`, see the behaviour change below.

---

### 3. `ColorSpaceDetailsByteConverter.Convert` 6-argument overload gained `RenderingIntent`

```csharp
- ColorSpaceDetailsByteConverter.Convert(details, bytes, bpc, width, height, image.Decode);
+ ColorSpaceDetailsByteConverter.Convert(details, bytes, bpc, width, height, image.Decode,
+                                        image.RenderingIntent);
```

The 5-argument overload is unchanged and now passes `RelativeColorimetric`.

---

### 4. `IResourceStore` gained four members

Only breaking if you implement the interface yourself. Minimal opt-out implementations:

```csharp
public IIccProfileService? IccProfileService => null;   // no ICC colour management

public IReadOnlyList<OutputIntent> DocumentOutputIntents => [];

public IReadOnlyList<OutputIntent> GetPageOutputIntents(DictionaryToken? pageDictionary)
    => [];

public IIccProfile? GetPageOutputIntentProfile(DictionaryToken? pageDictionary) => null;
```

Returning `null`/empty everywhere reproduces the previous behaviour exactly: ICC-based colour spaces fall back to their alternate, and no output intent is in effect.

---

### 5. `ICCBasedColorSpaceDetails` constructor gained `profile`

```csharp
  new ICCBasedColorSpaceDetails(numberOfColorComponents, alternateColorSpaceDetails, range, metadata,
+                               profile,
                                log);
```

Pass `null` to get the old behaviour: every conversion goes through `AlternateColorSpace`, and `BaseType` / `BaseNumberOfColorComponents` keep reporting the alternate's.

With a profile, `BaseType` stays `ColorSpace.ICCBased` and `BaseNumberOfColorComponents` is 3: the profile has placed the colour absolutely, in sRGB, so it is no longer a device colour for an output intent to reinterpret.

---

### 6. `CurrentGraphicsState` colour setters gained an optional profile

Source-compatible, binary-breaking (a recompile is enough).

```csharp
  void SetStrokingColor(ColorSpaceDetails colorSpace, double[]? operands,
+                       IIccProfile? outputIntentProfile = null);
  void SetNonStrokingColor(ColorSpaceDetails colorSpace, double[]? operands,
+                          IIccProfile? outputIntentProfile = null);
  void SetStrokingPatternColor(PatternColorSpaceDetails cs, NameToken name, double[]? operands,
+                              IIccProfile? outputIntentProfile = null);
  void SetNonStrokingPatternColor(PatternColorSpaceDetails cs, NameToken name, double[]? operands,
+                                 IIccProfile? outputIntentProfile = null);
```

The profile is retained rather than applied once, so a later `ri` re-manages the colour: a profile resolves a different transform per intent, which makes even a device colour vary by one. The `SetStrokingColor(IColor)` / `SetNonStrokingColor(IColor)` overloads are unchanged and still record a fixed colour that no intent change will move.

---

## Behaviour change independent of the ICC service

A `/Mask` stream that carries a `/ColorSpace` entry now throws `PdfDocumentFormatException` when `ParsingOptions.UseLenientParsing` is false. 8.9.6.4 says an image dictionary used as a stencil mask "shall not contain a ColorSpace entry", and the check was previously commented out with a TODO pointing at #1054. Lenient parsing, the default, is unaffected.

---

## Implementing `IIccProfileService`

Applying an output intent is impossible without a service to parse its profile, so the capability and the decision to use it live together:

```csharp
public bool TryGetProfile(ReadOnlyMemory<byte> profileBytes, out IIccProfile? profile) { /* ... */ }

// Convert device colours through the document's output intent profile? 14.11.5 says a processor is
// free to disregard an output intent, so this should be false unless previewing or proofing.
public bool UseOutputIntent => false;

// Which entry to prefer when a document declares several: an /S subtype, matched exactly.
// OutputIntent.PdfXSubtype is the usual answer; null is the built-in PDF/X, PDF/A, other order.
public string? PreferredOutputIntentSubtype => null;
```

Answering `false` and `null` reproduces the behaviour of a service that only parses profiles: `/ICCBased` colour spaces are colour-managed, device colours are not.

Malformed profiles are the implementation's problem: PdfPig hands over the bytes as the document wrote them. Returning `false`, or throwing, is safe - the profile is dropped and the colour space uses its alternate. What an implementation must not do is return a profile that reports success and then produces wrong colours.


BobLd:
For anyone interested, the work in progress implementation of the ICC transform is available in https://github.com/BobLd/PdfPig.Rendering.Skia/tree/feature/0.1.17-alpha/UglyToad.PdfPig.Rendering.Skia/Icc/Unicolour

Fix #1144

PR #1432, opened 2026-09-02
Configure TPageFactory<TPage> instance

adelobosko:
Currently, there is no way to access the factory instance after registering it with document.AddPageFactory<SKPicture, SkiaPageFactory>(); to configure additional options.

It relates to https://github.com/BobLd/PdfPig.Rendering.Skia/pull/146 when need to replace default font (`SKTypeface.Default` and use a font from file, https://github.com/BobLd/PdfPig.Rendering.Skia/issues/145).

### DIFF
--- a/src/UglyToad.PdfPig/Content/Pages.cs
+++ b/src/UglyToad.PdfPig/Content/Pages.cs
@@ -117,6 +117,35 @@
             pageFactoryCache.Add(type, pageFactory);
         }
 
+        /// <exception cref="InvalidOperationException"></exception>
+        internal TPageFactory GetPageFactory<TPage, TPageFactory>() where TPageFactory : IPageFactory<TPage>
+        {
+            Type type = typeof(TPage);
+            if (!pageFactoryCache.ContainsKey(type))
+            {
+                throw new InvalidOperationException($"Could not get page factory for page type '{type}' as it was not added.");
+            }
+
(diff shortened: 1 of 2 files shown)


BobLd:
@adelobosko I have not forgotten about that. I will review this weekend and get back to you. The reason it took so much time is that I'm not sure exposing the `TPageFactory<TPage>`s is the correct approach.

adelobosko:
@BobLd 

We also can add delegate / Action to `AddPageFactory` method. 
```csharp
internal void AddPageFactory<TPage, TPageFactory>(Action<TPageFactory>? factoryConfigurator = null) where TPageFactory : IPageFactory<TPage>
{
  // ....
 
  if (factoryConfigurator is not null)
      factoryConfigurator((TPageFactory)instance);

  AddPageFactory(pageFactory);
}
```

adelobosko:
@BobLd I have updated the pull

BobLd:
@adelobosko let me review shortly. I much prefer the approach of configuring once, then it's read only

BobLd:
@adelobosko this looks better. Do you mind adding a test covering the use of the delegate?

adelobosko:
@BobLd done, i have added a test

BobLd (approved):

BobLd:
@adelobosko I will run the pre-release pipeline shortly so a new NuGet package will be made available soon

PR #1433, opened 2026-09-02
Faster LZW decoding: copy sequences out of the output

KHDHaDi:
## What

Rewrites `LzwFilter` and its `BitStream`, and adds a benchmark and tests for them.

The decoder kept a `Dictionary<int, byte[]>` and copied the whole prefix for
every new code, the same layout as PDFBox's `LZWFilter`. Every new LZW code is
the sequence just written plus the first byte of the next one, and those bytes
already sit next to each other in the output. So the table now holds only a
start offset and a length per code, and decoding a code is a copy from earlier
in the output. The same idea drives giflib-turbo (Apache 2.0), which this was
checked against after the fact.

On top of that:

- The 256 literals are laid out ahead of the output so every code takes the
  same copy path.
- Sequences up to 16 bytes move as whole words via `BinaryPrimitives` into
  spare room kept past the output; the surplus is overwritten by the next code.
- Codes are read from a 64-bit word loaded at the current bit position, with
  no refill state.
- Buffers come from the `ArrayPool`, as in `BrotliFilter`, and the caller
  receives a copy of exactly the decoded length.

Everything is safe, span-based code and builds for all seven target frameworks.

## Numbers

BenchmarkDotNet, .NET 8, the sixteen LZW streams of `ssm2163.pdf` (230 KB
compressed, 560 KB decoded), baseline is the 0.1.17-alpha NuGet package:

| Benchmark      | Before      | After    | Allocated before | Allocated after |
|----------------|-------------|----------|------------------|-----------------|
| All 16 streams | 9.3–12.7 ms | 0.76 ms  | 25.2 MB          | 0.54 MB         |
| Largest stream | 2.0–2.7 ms  | 0.22 ms  | 5.4 MB           | 0.15 MB         |

Variants that were measured and dropped: masked table indices, a word loop
for sequences up to 64 bytes, and a 4-byte refill of the previous bit
accumulator. None gave a reliable gain.

## Verification

- `LzwFilterTests`: the ISO 32000-1 worked example, round trips through an
  in-test encoder for both `EarlyChange` values covering 12-bit codes and a
  table reset, the PNG predictor path, an invalid code, and the SHA-256 of
  the LZW streams in `ssm2163.pdf` and `byz.pdf` as decoded before the change.
- Byte-identical to the previous decoder on all 426 LZW streams found in a
  private corpus of 866,944 PDFs (603 MB decoded). 74 streams with a predictor
  could not be compared by that tool and are covered by the unit test instead.
- Existing filter integration tests pass on net8.0 and net471.

## Behaviour change

Data that ends without an EOD marker now yields what was decoded so far,
as PDFBox does, instead of throwing and losing the whole stream. A code that
is not in the table and has no preceding sequence still throws.


KHDHaDi:
@BobLd : Sorry to bother you again, but this PR is also the foundation for faster PNGPredictor and decoding. Gives 12x LZW speed... But I don't mind to wait. Don't want to make to much work for you,  but I have too much fun optimizing the performance :-)

BobLd:
@KHDHaDi no problem at all! performance improvements are more than welcome, keep up the fun!

BobLd (approved):

Issue #1404, opened 2026-08-21
FlateFilter hands back the still-compressed input as decoded content, and discards the exception that says so


KHDHaDi:
# FlateFilter hands back the still-compressed input as decoded content, and discards the exception that says so

## What this is not

This is **not** a request to revisit #1254. Verifying the zlib checksum broke real
documents (#1235), the revert to lenient processing was right, and nothing below
asks for it back. Leniency is not the problem.

The problem is that the caller cannot tell leniency apart from success.

## Summary

`FlateFilter.Decode` wraps the whole decompression in a bare catch and returns the
untouched input:

```csharp
try
{
    ...
    return Decompress(input, predictor, colors, bitsPerComponent, columns);
}
catch
{
    // ignored.
}

return input;
```

`Decompress` one frame below does the right thing — it raises a dedicated, public
exception type:

```csharp
catch (InvalidDataException ex)
{
    throw new CorruptCompressedDataException("Invalid Flate compressed stream encountered", ex);
}
```

`UglyToad.PdfPig.Fonts.CorruptCompressedDataException` is `public`. The library
models this condition, constructs the object, and then the bare catch throws the
knowledge away. No log, no flag, no return value distinction.

The consequence is not "less data than hoped". It is that a corrupt stream returns
**the compressed bytes** to the caller, in the position where decoded content
belongs.

## Measurements

PdfPig 0.1.15, .NET 10, Release. `stream.Decode(DefaultFilterProvider.Instance)`
over a `StreamToken` declaring `/Filter /FlateDecode`. The reference payload is
4,096 bytes of `'A'`, zlib-compressed to 28 bytes.

| Input | bytes in | bytes out | output identical to input? |
|---|---:|---:|---|
| valid stream | 28 | 4,096 | no — correct |
| ASCII text, not deflate at all | 27 | 27 | **yes — raw input returned** |
| eight zero bytes | 8 | 8 | **yes — raw input returned** |
| valid stream, truncated in half | 14 | 0 | empty, no signal |
| valid stream, one byte flipped mid-stream | 28 | 28 | **yes — compressed bytes returned as content** |
| valid stream, last byte of the Adler-32 flipped | 28 | 4,096 | no — correct data, checksum ignored |

The last row is #1254 working as intended and is fine. Rows 2, 3 and 5 are the
concern: the return value is the input, and nothing distinguishes it from a
successful decode of content that happens to look like that.

[...]

PR #1434, opened 2026-09-03
Faster predictors and Flate: decode in place, SIMD Sub and Paeth, 64 KB reads

KHDHaDi:
## What

Rewrites `PngPredictor` and the plumbing of `FlateFilter`, adds benchmarks for
both, and takes the empty-on-failure behaviour from #1425 without its log and
leniency switch. Eight commits, each standing on its own.

### Predictors

The predictor was a stream in front of the output: every row copied into a
row buffer, decoded against a copy of the previous row, copied on into a
MemoryStream. Now:

- Rows are decoded where they lie, the row above them being the reference,
  and the PNG filter type bytes are dropped as the rows move up. Nothing is
  allocated per stream beyond one zero row.
- The Flate filter hands each inflated block to the decoder while it is still
  in the cache. A separate pass over the finished data was measured slower
  than the old stream and is not what this does.
- For images the stream dictionary gives the height, so the result is
  allocated at its final size and rows are decoded straight into it from a
  small recycled input buffer; no move, no copy afterwards. A wrong height is
  grown around or trimmed.
- Sub and Up read residuals where they lie and write the decoded row where it
  goes, one pass instead of two.
- Sub is undone as a vector prefix sum on .NET 7 and later (shift by one,
  two, four pixels and add, carry into the next vector), the approach of
  connorskees/simd-png and the SIMD prefix-sum literature, using the byte
  shuffle instruction directly since `Vector128.Shuffle` with computed
  indices is not accelerated.
- Paeth for three and four byte pixels works in 16-bit lanes as libpng,
  wuffs and ImageSharp 2.1 do; the scalar Paeth picks its prediction with
  masks instead of branches. Sub and Average carry the left pixel in locals
  instead of reading back the byte just stored; Up adds whole vectors.

Everything is safe, span-based code. The SIMD paths need .NET 7 or later on
x86 with SSSE3 or on ARM64; every other target takes the scalar paths, which
are two to four times faster than the release on their own.

### Flate

- Inflates straight into a rented buffer and hands the caller a copy of the
  finished length, as `BrotliFilter` does.
- Reads 64 KB at a time. Every read is a call into native zlib that leaves
  and re-enters its fast path; with zlib-ng on .NET 9 the inflating is fast
  enough for 8 KB reads to cost a sixth of the total. A stream found damaged
  is inflated once more in 4 KB reads so that everything up to a few
  kilobytes before the damage is kept, as before and as PDFBox does.
- When decoding fails outright the filter returns an empty result instead of
  handing the compressed input back as though it were content (#1404,
  #1425). Header and checksum stay unchecked, as in PDFBox's
  `FlateFilterDecoderStream`.

## Numbers

BenchmarkDotNet, .NET 8, baseline is the 0.1.17-alpha package.

| Benchmark                                      | Before   | After    |
|------------------------------------------------|----------|----------|
| Predictor alone, 33 real Sub images, 8 MB      | 4.6 ms   | 0.75 ms  |
| Predictor alone, RGB Sub / Up / Average / Paeth | 1.9 / 1.2 / 2.8 / 21.9 ms | 0.36 / 0.23 / 1.4 / 4.8 ms |
| Predictor alone, corpus mix (59 % None, 36 % Sub, 4 % Up) | 1.1 ms | 0.32 ms |
| Flate with predictor, 33 real images           | 16.1 ms, 20.7 MB | 11.9 ms, 7.7 MB |
| Flate, 1,786 real content and font streams     | 50 ms    | 45 ms (.NET 8), 29 ms (.NET 9) |

The bare inflater takes 41 ms on .NET 8 and 24 to 27 ms on .NET 9 for those
1,786 streams; the filter is within 10 percent of it.

## Verification

- `PngPredictorTests`: the previous implementation kept as oracle over 18
  parameter sets including padded last rows; the incremental, in-place and
  separate-output decoders against whole-buffer decoding with random block
  sizes; the content of 49 real document streams by SHA-256.
- `FlateFilterTests`: damage part way keeps the first half intact; a failure
  yields nothing.
- Byte-identical to the previous filters on 420,368 Flate streams from a
  sample of 173,389 documents (161 GB decoded), 43,631 of them with
  predictors covering every parameter set found; decoding in 49 percent of
  the previous time on .NET 8.
- Full suite green on net8.0; filter tests on net471.

## Behaviour changes

- A stream that fails to decode yields an empty result instead of its
  compressed bytes.
- Absurd predictor parameters (row length beyond an int) are refused instead
  of wrapping around; a zero-byte TIFF row no longer loops forever.

## Notes

- Techniques were taken from libpng (libpng licence), wuffs (Apache 2.0 or
  MIT), connorskees/simd-png (MIT) and ImageSharp 2.1.9 (Apache 2.0); no code
  was copied. ImageSharp 3.x is under the Six Labors Split License and was
  not used.
- `LzwFilter` is touched at its predictor call; that conflicts with the LZW
  branch in three lines.
- The benchmark project now targets net8.0 and net9.0, so `dotnet run` needs
  `-f net8.0`.


BobLd:
@KHDHaDi the PR has conflicts, do you mind resolving them?

KHDHaDi:
@BobLd: done

BobLd:
@KHDHaDi I will need a bit of time as the PR is quite involved, but looks good so far.

Side question (no code change required, it's more out of curiosity): I see the SIMD changes use `Vector128`, why fixing that on 128? Is this because this is the most common?

BobLd:
@KHDHaDi below is a code review done via AI - as a general rule, never fully trust pdf inputs when it comes to size of array (e.g. `/DL`).


### 1. An untrusted `/DL` can force a multi-gigabyte up-front allocation

`src/UglyToad.PdfPig/Filters/DecodeBuffer.cs:54`

The guard accepts a stated length up to `compressedLength * 4096`, then rents exactly that much
before a single byte is inflated. `/DL` is spec'd as a hint only (ISO 32000-1, 7.3.8.2) and is
attacker- or corruption-controlled.

A 1 MB compressed stream declaring `/DL 2000000000` passes `2e9 <= 1e6 * 4096` and rents ~2 GB.
Before this change the same stream started at 4 MB and grew only as real data arrived, so this is a
roughly 500x up-front amplification driven purely by an unverified dictionary number.
`FlateFilter.Decode`'s blanket `catch` then swallows the resulting `OutOfMemoryException` and returns
`Memory<byte>.Empty`, so the spike is invisible to callers and to a test run.

The comment at `DecodeBuffer.cs:16-19` names the right number and then does not use it: deflate's
worst case is 1032:1, not 4096:1.

**Suggested fix:** tighten `MaximumExpansion` to deflate's actual bound, and add an absolute ceiling
(a few tens of MB) above which the stated length is ignored and the `factor`-based growth path is
used instead. Streams that legitimately state a length — embedded files, font programs — are nowhere
near such a ceiling, so the optimisation keeps its value.

### 2. The overflow guard in `CalculateRowLength` is defeatable

`src/UglyToad.PdfPig/Filters/PngPredictor.cs:96`

The comment promises absurd parameters are "refused rather than turned into a negative or a tiny
row", but `((long)columns * colors * bitsPerComponent)` still overflows. `colors` is clamped to <= 32
(`FlateFilter.cs:70`, `LzwFilter.cs:69`, and the same line in `BrotliFilter`) but never to >= 1, and
`bitsPerComponent` is not validated at all.

Concrete input: `Predictor 12, Colors 32, BitsPerComponent 1073741824, Columns 536870912`.
`536870912 * 32 = 2^34`, times `2^30` is exactly `2^64`, which wraps to 0, so `rowLength = 0` and
neither the `< 0` nor the `> int.MaxValue` check fires. The stream then decodes silently to nothing
instead of throwing. `Colors 0` reaches the same place more simply.

The same unvalidated values feed `bytesPerPixel = ((colors * bitsPerComponent) + 7) / 8` at
`PngPredictor.cs:157` and `:926`, which is plain `int` arithmetic and can wrap negative — that one
throws out of `Span.Slice`, which `LzwFilter` and `BrotliFilter` do not catch.

No memory-safety consequence (a wrapped `rowLength` still lands in `[0, int.MaxValue)` and is used
consistently downstream), but the guard does not do what it says.

**Suggested fix:** validate at the filter entry — `colors` in `[1, 32]`, `bitsPerComponent` in
`{1, 2, 4, 8, 16}` as the spec allows. That kills the whole class in one place and is cheaper than
guarding each arithmetic site.

### 3. Buffer growth copies the whole old array, not the written prefix

`src/UglyToad.PdfPig/Filters/LzwFilter.cs:264`, `src/UglyToad.PdfPig/Filters/FlateFilter.cs:333`

Both `EnsureCapacity` and `EnsureOutput` do `output.AsSpan().CopyTo(grown)`.

- In `LzwFilter` the rented buffer can be far larger than `position` (ArrayPool rounds up, and the
  initial rent is `/DL`-sized), so each growth copies pool garbage past the live data.
- In `FlateFilter.EnsureOutput` only `decoder.DecodedLength` bytes are meaningful; the rest is the
  uninitialized tail of `AllocateResult`, copied for nothing.

`FlateFilter.Grow` (`:338`) right next door already takes a `length` parameter and does this
correctly. Worth making the other two match — the fix is to thread the written length through.

### 4. `compact: false` combined with `RestartInput()` would silently mis-decode

`src/UglyToad.PdfPig/Filters/PngPredictor.cs:330`

The non-compact path detects the first row with `ConsumedLength == 0`; the compact path uses
`DecodedLength == 0` (`:370`). No caller mixes `compact: false` with `RestartInput()` today, so this
is latent — but nothing says the two are incompatible, and `RestartInput`'s doc reads as generally
applicable. A caller that combined them would decode each restarted batch's first row against
`zeroRow` instead of the row above, producing wrong output with no error.

**Suggested fix:** either document the exclusion on the `compact` constructor parameter, or use
`DecodedLength == 0` in the non-compact path too — equivalent for current usage and
correct-by-construction.

### 5. `MaximumCapacity` is now declared four times

`DecodeBuffer.cs:13` (already `public`), `FlateFilter.cs:53`, `LzwFilter.cs:55`,
`BrotliFilter.cs:64`. The latter three can be dropped in favour of `DecodeBuffer.MaximumCapacity`.

### 6. Test coverage gap: two `SubVectorized` pixel widths

`src/UglyToad.PdfPig.Tests/Filters/PngPredictorTests.cs:102`

`ParameterSets` exercises `bytesPerPixel` 3, 4 and 6 through the vectorised `Sub`, but not 2 or 8.
8 is `SubMasks.MaxBytesPerPixel` exactly — the boundary where `block == Vector128<byte>.Count` and
the carry mask is widest, i.e. the case most likely to be wrong.

Adding `{ 15, 1, 16, 40 }` (bpp 2) and `{ 15, 4, 16, 12 }` (bpp 8) closes it; both flow through the
existing `LegacyPredictor` oracle for free.

KHDHaDi:
> @KHDHaDi I will need a bit of time as the PR is quite involved, but looks good so far.
> 
> Side question (no code change required, it's more out of curiosity): I see the SIMD changes use `Vector128`, why fixing that on 128? Is this because this is the most common?

Not really because it's the most common, though that helps. Three reasons, and one place where the code does go wider.

One pixel is 128 bits. The Sub and Paeth filters work like a chain: each pixel depends on the pixel to its left, so you can only finish one pixel at a time. What you can do in parallel is the bytes inside one pixel. A pixel here is at most 8 bytes (4 colours of 16 bits), and the arithmetic needs 16-bit slots so nothing overflows: 8 × 16 = 128 bits. A 256-bit register would hold two pixels, but the second has to wait for the first, so nothing is gained.

The instruction the trick needs is 128 bits wide. Decoding Sub fast means shifting a block by one pixel and adding, a few times over. That byte-wise reshuffle is a 128-bit instruction on both x64 (pshufb) and ARM (tbl). The 256-bit version on x64 works in two separate halves that can't exchange bytes, which breaks the trick.

128 bits works everywhere. Every x64 and every ARM64 CPU runs Vector128 in hardware. Vector256 needs AVX2 on x64 and doesn't exist on ARM, where .NET would emulate it in software and be slower. A 256-bit kernel would need a 128-bit fallback anyway.

Where wider helps, the code uses it. The Up filter just adds one row to the row above, no pixel waits for another, so it uses Vector<byte>, which the runtime sizes to the machine: 256 bits on AVX2, 128 elsewhere.

Where the tricks come from. The Sub shift-and-add is the parallel prefix sum from Hillis and Steele, 1986, applied to PNG rows by Google's Wuffs, the Rust project simd-png and ImageSharp 2.1; ours is closest to simd-png because it uses a shuffle with a mask table per pixel width, which is what handles 3 and 6 bytes per pixel with one piece of code. The branchless Paeth, computing all three candidates and blending them with masks instead of ifs, comes from libpng's Intel SSE2 code by Mike Klein at Google, 2016, later ported to ARM inside libpng and used by Wuffs, Skia, ImageSharp and the Rust png crate. Only the ideas were taken, not code; the licences (libpng, Apache 2.0, MIT) are all compatible, and the comments in the code credit the sources.

KHDHaDi:
@BobLd : fixed everything from the review and some new ones I found on the way

BobLd:
@KHDHaDi one last review, in `PngPredictor`:

> The old filters clamped (`colors = Math.Min(..., 32)`) and accepted any `BitsPerComponent`; the new shared `Parameters`/`Decoder` path validates and throws. 

This is a tolerance regression, i.e. pdf docs that used to be readable and would not anymore. The check is correct, but can we clamp before the check? I believe parsing options and the lenient parsing flag do not reach this part of the code (**let's not change that for the moment**), so let's be lenient by default and clamp (maybe put a short comment in the code where it happens)

EDIT: I am going to merge as is and fix in a PR

PR #1440, opened 2026-09-05
Scan for xref keywords with a byte window over blocks

KHDHaDi:
XrefBruteForcer, which runs when the xref chain cannot be read, added every byte to a ring buffer and asked it four times whether a keyword had just ended, each answer a loop over the keyword; every number was gathered in a list, turned into a string and parsed. The last eight bytes now sit in a ulong, whitespace normalized, and the byte just read selects the one masked comparison that could match, as the four keywords end in three different bytes. Numbers accumulate in a long, with overflow treated as no number, like long.TryParse did. The input is read in 64 KB blocks and scanned from the array instead of through IInputBytes a byte at a time; after a keyword hands the input to another parser the scan resumes where that parser left it, as before. On 13 documents that need the scan (3.5 MB on average) opening takes 9 ms instead of 32. A signature of trailer, page text, letter and image counts is unchanged for 21,673 documents of a corpus sample, 21 of them brute-forced.

### DIFF
--- a/src/UglyToad.PdfPig/Parser/FileStructure/XrefBruteForcer.cs
+++ b/src/UglyToad.PdfPig/Parser/FileStructure/XrefBruteForcer.cs
@@ -89,34 +89,43 @@ internal static class XrefBruteForcer
             numberOverflowed = false;
         }
 
-        var block = ArrayPool<byte>.Shared.Rent(BlockSize);
+        // The block has one byte of room in front of the data read into it, for the byte that
+        // the previous read ended on: it is visited only once the next read has shown that it
+        // was not the last byte of the input. That way the scan needs no length from the input,
+        // which a buffering stream does not know until it has been read to the end, and it
+        // leaves the last byte unvisited like the byte-wise loop it replaces. No keyword that
+        // ends on the last byte could yield anything: an object with nothing after it, a table
+        // or a trailer with nothing left to read.
+        var block = ArrayPool<byte>.Shared.Rent(BlockSize + 1);
 
         try
         {
-            // The scan visits every byte but the last one, as the byte-wise loop it replaces did.
             // Positions are computed from the index: the input reports an offset one past the
             // byte it is on. Where a keyword hands the input to another parser, the scan resumes
-            // from wherever that left the input, again as before.
-            var last = bytes.Length - 1;
+            // from wherever that left the input, as the byte-wise loop did.
             var next = 0L;
+            var hasPending = false;
 
-            while (next < last)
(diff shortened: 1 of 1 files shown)


KHDHaDi:
What the brute-force scan does. When a PDF's cross-reference table is broken, PdfPig has to find the objects itself. It walks the whole file byte by byte, looking for four keywords: obj, xref, /XRef and trailer, and it remembers the two numbers before each obj, because those are the object number and generation.

Why it was slow. For every single byte, the old code pushed the byte into a small ring buffer and then asked that buffer four separate questions: "does it end in  obj?", "in  xref?", "in /XRef?", "in trailer ?". Each question was a loop comparing character by character, with a modulo per character to find the position in the ring. On top of that, every number found in the file was collected into a list, turned into a string and parsed with long.TryParse. And every byte was fetched through three interface calls on IInputBytes. So a 3.5 MB file cost about 32 ms just for this scan.

Step one: one integer instead of a ring buffer. The last eight bytes are kept in a single ulong. Reading a new byte is one shift and one OR. A keyword has been read when the low bytes of that integer equal the keyword's packed bytes, which is one masked comparison. Better still, the four keywords end in three different bytes: j, f and a space. So the byte just read tells you which comparison could possibly match, and most bytes need none at all. Numbers are added up directly as a long while the digits pass by, with an overflow flag standing in for a failed TryParse.

Step two: read in blocks. Instead of asking the input for one byte at a time, the scan reads 64 KB into an array and runs over the array. Positions are computed from the index. When a keyword hands control to another parser, which may move the shared input, the scan simply resumes from wherever that parser left the input, exactly as the old loop did.

What stays the same. Every byte is visited in the same order, whitespace is normalized the same way, comments are skipped the same way, the same numbers are queued, and the last byte of the file stays unvisited as before. The only difference is how fast each byte is handled.

Result. 32 ms down to 9 ms per document that needs the scan, about 3.5x. The signature of trailer, page text, letters and images is identical to the original on more than 21,000 corpus documents.

One lesson from CI. My first block version took the file length from the input to know where to stop. A buffering stream only knows how much it has read so far, so the scan saw almost nothing. The fix keeps the last byte of each block back until the next read proves it was not the end of the file. No length needed.

BobLd:
Note for later, regarding some `Seek` that could be skipped:
> Only the table/stream branches below leave the input positioned somewhere other than `next`; the trailer branch and a plain block read both leave it exactly there, so seeking again just cost a redundant read for no change in position.

BobLd (approved):

PR #1441, opened 2026-09-05
Compute the deep token hashes on first use

KHDHaDi:
DictionaryToken, ArrayToken and StreamToken have hashed themselves in the constructor since d67d7928 introduced the resource and form caches that key on them: every entry, and for a stream every byte, on every object parsed. Only a token used as a dictionary key ever needs the value. It is now computed on the first GetHashCode call and kept, with the flag written and read through Volatile so that no thread sees the flag before the value.

Opening a document and parsing every object it has takes 6 to 8 % less time on a corpus sample; the parsed objects are unchanged for 2,592 documents.

### DIFF
--- a/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
+++ b/src/UglyToad.PdfPig.Tokens/ArrayToken.cs
@@ -1,6 +1,7 @@
 ﻿namespace UglyToad.PdfPig.Tokens
 {
     using System;
+    using System.Threading;
     using System.Collections.Generic;
     using System.Text;
     using Core;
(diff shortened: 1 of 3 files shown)


BobLd:
@KHDHaDi can you remove the comment `        // The hash is deep, over every entry, and is asked for only where a token serves as a
        // dictionary key. Computing it in the constructor made every parsed object pay for it;
        // it is computed on first use instead.`

There is no need to refer to the previous behaviour, the code is self-explainatory

KHDHaDi:
@BobLd : Yes, good to know your prefered style. I'm always between does somebody else get the idea of what I want to achive and why. And on the other hand, yes, this code is self-explainatory and easy to understand. And well AI does it in 60% wrong, does not explain the evil to understand parts, but does comment something like this all the time... well trained by humans.

KHDHaDi:
> @BobLd : Yes, good to know your prefered style. I'm always between does somebody else get the idea of what I want to achive and why. And on the other hand, yes, this code is self-explainatory and easy to understand. And well AI does it in 60% wrong, does not explain the evil to understand parts, but does comment something like this all the time... well trained by humans.

done

